PR #5, opened 2015-06-29
fixed apt-get command for ubuntu in README.md

skorasaurus:
Thanks Richard for making this. 

I just added a missing install parameter on the apt-get instructions.


systemed:
Good spot, thank you!

PR #18, opened 2015-09-24
Random improvements

zerebubuth:
Apologies for this random grab-bag of improvements (if you even consider all of them to be improvements...)

The main changes:

* Compiles sources to `.o` temporaries as an intermediate step. This means changes in one `.cpp` file don't require recompiling everything and leads to quicker recompiles. Also, `-j` should now provide some speedup for parallel builds.
* Uses `std::unordered_map` to store nodes rather than `std::map` since ordered access isn't required. This gives a modest speedup.
* A few attempts to avoid (re-) allocation of arrays.
* Added a vagrant config to make creating a compatible VM with the dev environment pre-installed easier.
* Added a git ignore file so that it's easier to see what the important changes are.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,13 @@
+# object files created during the build and the output binary itself
+*.o
+tilemaker
+
+# editor temporaries
+*~
+
+# vagrant local stuff
+.vagrant/
+
+# protocol buffers generated headers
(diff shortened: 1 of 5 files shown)


systemed:
That's superb - no apologies needed. Thank you.

PR #20, opened 2015-10-05
Support to show traceback of Lua runtime error.

uievolutionkk:
Modified to show Lua traceback when runtime error occurs in process.lua.


systemed:
Thank you for both of these!

PR #22, opened 2015-10-05
Modified to be able to set compress option (gzip/deflate) in config.json

uievolutionkk:
Modified to specify type of compression in config.json.
Because mapbox vector tile renderer (https://github.com/mapbox/mapbox-gl-native/) 
handles deflated compression only.

NOTE: "compression" value in config.json has been changed from boolean to string.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -27,7 +27,7 @@ It also includes these global settings:
 * `maxzoom` - the maximum zoom level at which any tiles will be generated
 * `basezoom` - the zoom level for which Tilemaker will generate tiles internally (should usually be the same as `maxzoom`)
 * `include_ids` - whether you want to store the OpenStreetMap IDs for each way/node within your vector tiles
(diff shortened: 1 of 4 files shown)


systemed:
Interesting. Mapbox GL Native does appear to support gzipped tiles when served with a 'Content-Encoding: gzip' header - that's how I've been using `resources/mbtileserver.rb`. But this is a good change anyway, so merging!

tinoue:
Maybe it is an implementation issue of mapbox-gl-native. In https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/util/compression.cpp  , decompress() uses inflateInit() which supports deflate only.

systemed:
Ah, that starts to make sense. Presumably iOS decompresses the data (based on the Content-Encoding tile) before passing it across to Mapbox GL, whereas if the tile arrives 'as is' then compression.cpp can inflate it.

PR #24, opened 2015-10-10
Support building on Windows

alex85k:
Thank you very much for this good tool!

Usage of C++11 allows simple cross-platform builds. I had to do only one small change (define uint type) to make it compile under Visual Studio 2015.

It would be useful to add cmake build scripts for systems that have libraries in different locations.
For example, osrm and libosmium have them.

cmake will not break the existing Makefile build (special check is included)

The script should also work on Linux/BSD/Macos but this needs additional checking. Commands to build are something like
```bash
git clone https://github.com/alex85k/tilemaker.git
#if merged, git clone https://github.com/systemed/tilemaker.git :)
cd tilemaker
mkdir build && cd build
cmake .. -G "Unix Makefiles"
make
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,52 @@
+cmake_minimum_required(VERSION 2.8)
+
+project(tilemaker)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+
+if("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}")
+  message(FATAL_ERROR "In-source builds are not allowed, please use a separate build directory.")
+endif()
+
+include_directories(include)
+include_directories(${CMAKE_BINARY_DIR}) # for generated files
+
+find_package(Boost 1.56 REQUIRED COMPONENTS system filesystem program_options)
+include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
+
+find_package(Protobuf REQUIRED)
+include_directories(${PROTOBUF_INCLUDE_DIR})
+
+find_package(SQLite3 REQUIRED)
+include_directories(${SQLITE3_INCLUDE_DIR})
+
+find_package(libshp REQUIRED)
+include_directories(${LIBSHP_INCLUDE_DIR})
+
+find_package(Lua REQUIRED)
+include_directories(${LUA_INCLUDE_DIR})
+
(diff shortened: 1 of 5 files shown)


pnorman:
would this mean there are two build systems set up?

alex85k:
Actually, yes. Makefile will be for "simple classical Linux systems", where all libraries are placed correctly and CMake is for Windows and non-standard Linux installations (like non-root installed home-made libraries). The build process is simple enough and the support of CMakeLists.txt should be easy.

alex85k:
Re-pushed after checking on FreeBSD 10.2 - added minimal Boost version 1.56, -std=c++11 option and ZLIB in capital letters.
Then the build was successful (had to build boost 1.59 from source and build luabind on top of it).
Then
``
mkdir build && cd build && cmake .. && make install
``
worked perfectly.
Makefile needed some simple manual pathing (g++->clang++, lua library names), then it worked too.

systemed:
Looks good to me - I expect it'll be a minority interest and that the standard makefile will be enough for most people, but if it helps people I don't see why not. I'll probably ping you to update it if we ever add other dependencies though. :)

alex85k:
Thank you!
I agree that this was not a high priority feature, but it is still convenient enough.
For example, on OpenSuse 14.1 beta there are errors ``cannot find -llua5.1, cannot find -lluabind` and cmake finds them automatically.

Can change scripts anytime if needed. :)

systemed:
Good stuff - thanks for the PR!

ergoen:
Sorry for commenting on an old issue, but cmake builds for me are failing unless I replace the following line in the CMakeLists file
"add_executable(tilemaker vector_tile.pb.cc osmformat.pb.cc src/tilemaker.cpp)"
with
"add_executable(tilemaker vector_tile.pb.cc osmformat.pb.cc clipper/clipper.cpp src/mbtiles.cpp src/pbf_blocks.cpp src/coordinates.cpp src/osm_store.cpp src/helpers.cpp src/output_object.cpp src/read_shp.cpp src/read_pbf.cpp src/osm_lua_processing.cpp src/write_geometry.cpp src/shared_data.cpp src/tile_worker.cpp src/tile_data.cpp src/osm_mem_tiles.cpp src/shp_mem_tiles.cpp src/tilemaker.cpp)"

I also have to add the corresponding uint redefinition in a few more places to make it compile on windows. Has something changed (during the last few years :P) in cmake?

PR #27, opened 2015-12-03
Make SQLite output fast, by unifing transaction and use asynchronous writting.

grafi-tt:
I made SQLite output fast, by unifing transaction and use asynchronous writting.

This technique is described in following.
http://stackoverflow.com/questions/1711631/improve-insert-per-second-performance-of-sqlite


systemed:
Superb! Thank you.

PR #28, opened 2015-12-10
Make the global data store, and make it available geometry data when processing relations

grafi-tt:
Currently OSM IDs for node, way and relations are handled in really ad-hoc manner. This patches fixed this by creating a globally used data structure. This change also make it easy, optimization of internal data structure for efficient memory use.

And the flow of reading .pbf is changed, to make the information about coordinates of nodes available when processing relations.

I added functions to be used by Lua: `IsClosed` and `Area`. `Area` function can be used for multipolygons.

### DIFF
--- a/src/osm_object.cpp
+++ b/src/osm_object.cpp
@@ -20,11 +20,12 @@ class OSMObject { public:
 	map< string, RTree> *indices;			// Spatial indices
 	vector<Geometry> *cachedGeometries;		// Cached geometries
 	map<uint,string> *cachedGeometryNames;	// Cached geometry names
-	node_container_t *nodes;				// Node storage
+	OSMStore *osmStore;						// Global OSM store
 	bool isWay, isRelation;					// Way, node, relation?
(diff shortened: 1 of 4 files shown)


grafi-tt:
I tested that these changes do not affects current outputs in my usecase (a full map of Hawaii).

systemed:
This looks great - thank you!

Are there any impacts on memory usage that you know of?

grafi-tt:
I suppose there is no impact now, because the internal data structures is mostly unchanged. The only difference is: responsibility to store innerWays and outerWays is moved from each OutputObject to RelationStore. This change should have little impact.

Checking all ways in any relation before processing relations costs some memories, but collected data are cleared when relations are processed, and don't affect maximal memory usage.

systemed:
Tested it with my (moderately complex) Lua and config files, and it works fine. Thanks again!

grafi-tt:
Thank you!

PR #29, opened 2015-12-18
Shapefile fix

grafi-tt:
Fix the handling of shapefile.

Note that `assign_points` makes the size of interior_rings 0. Therefore subsequent appending into each interior_ring has have no effect.


systemed:
Thanks!

PR #30, opened 2015-12-18
Added Lua functions - init_function() and exit_function()

grafi-tt:
Two functions `init_function` and `exit_function` are called if available. It turned out these functions are useful, to setup things and output statistic informations.


systemed:
Good idea. (I've wanted this in OSRM for a while!)

PR #35, opened 2016-02-12
Add missing dependency to vagrant file

rory:
libshp-dev is required to compile tilemaker. It's in the docs.


systemed:
Thank you!

Issue #25, opened 2015-11-04
Suggestion: adaptive simplify level


tinoue:
> ```
> "low_roads": { "minzoom": 9, "maxzoom": 11, "write_to": "roads", "simplify_below": 12, "simplify_level": 0.0001 }
> ```

In my understanding, the above configuration applies the same simplfy_level 0.0001 to each zoom level below 12.
Though it would be helpful to apply simplfy level 0.0002 to zoom level 11, 0.0004 to zoom level 10 and so on.

systemed:
Agreed, that's a good idea.

PR #38, opened 2016-02-18
Actual length and adaptive simplification

grafi-tt:
We implemented adaptive simplify level, that is proposed in the issue #25.
The ratio can be specified by the `"simplify_ratio"` key in the config json file.
Its default is `1.0`, so the existing behavior would not change.

We also implemented `"simplify_length"`, that specifies simplify level by actual length (in meters).
It is useful because  `"simplify_level"` produces inconsistent output when
a country ranging over broad latitudes is processed.

In addition, `Length()`, `ScaleToMeter()`, and `ScaleToKiloMeter()` are added to Lua.
They are easily implemented.

Here we propose version change. So `CHANGELOG.md` is updated in the last commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,23 @@
 # Changelog
 
+## [] - 2016-02-XX
+
+### Added
+- Add cmake scripts and support MSVC on Windows (@alex85k)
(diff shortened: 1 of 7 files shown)


systemed:
This is really good - many apologies for not merging it for so long!

PR #43, opened 2016-05-28
Support shapefile polygons with multiple exterior rings

### LINKED ISSUE #42: Support shapefile polygons with multiple outer rings
Firstly, thanks for making this great tool!

The [ESRI Shapefile Technical Description](https://www.esri.com/library/whitepapers/pdfs/shapefile.pdf) states that a shapefile polygon may contain multiple outer rings, in effect representing a multi-polygon geometry. 

Tilemaker currently assumes all shapefile polygons consist of only 1 outer ring. Therefore, any shapefile polygon containing multiple outer rings will fail Boost's geometry checks on ring orientation. An example of such a dataset where this occurs is [Ordnance Survey Boundary-Line](https://www.ordnancesurvey.co.uk/business-and-government/products/boundary-line.html).

I couldn't find anything in the boost geometry docs for organising a set of rings into multiple polygons. It appears that OGR has code for [handling this](http://www.gdal.org/classOGRGeometryFactory.html#afa3e76499553c3dc0a1621af98b88be6), but it doesn't use boost geometry and in any case is a bit too heavyweight a dependency to add just for this purpose. 

Therefore, it would be nice to add a bit of code to tilemaker to support this, possibly making some assumptions about the ring ordering/overlaps in the input so that heavy spatial calculations can be avoided when reading shapefiles - something along the lines of how OGR does it.

I'd be happy to make a PR and give this a try.

fofanov:
Resolves #42 by adding support for shapefile polygons that represent a multi-polygon geometry. 

To avoid expensive computations in reconstructing the multi-polygon from its rings, we make the assumption that the ring ordering has been pre-processed such that each polygon's exterior ring is immediately followed by its interior rings in sequence.

### DIFF
--- a/src/read_shp.cpp
+++ b/src/read_shp.cpp
@@ -97,7 +97,7 @@ void readShapefile(string filename,
 
 	for (int i=0; i<numEntities; i++) {
 		SHPObject* shape = SHPReadObject(shp, i);
-		int shapeType = shape->nSHPType;	// 1=point, 3=polyline, 5=polygon [8=multipoint, 11+=3D]
+		int shapeType = shape->nSHPType;	// 1=point, 3=polyline, 5=(multi)polygon [8=multipoint, 11+=3D]
 	
 		if (shapeType==1) {
 			// Points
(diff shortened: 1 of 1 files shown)


systemed:
Excellent - thank you!

Issue #41, opened 2016-04-19
Create two objects in same layer from the same way


ThibaudM:
Hello,

I tried to use the following in lua:

```
function way_function(way)
    way:Layer("layer1", true)
    way:Attribute("key", "value1")

    way:Layer("layer1", true)
    way:Attribute("key", "value2")
end
```

This code generates only one feature (instead of 2) in the final geojson.
#### Finding the problem

In the source code this behaviour is due to:
`map< uint64_t, unordered_set<OutputObject> > tileIndex;` in _tilemaker.cpp_ and

```
bool operator==(const OutputObject& x, const OutputObject& y) {
    return (x.layer == y.layer) && (x.objectID == y.objectID) && (x.geomType == y.geomType);
}
```

in _output_object.cpp_ because there is no verification of equality for attributes field.
#### Trying to resolve

For me there are 2 solutions:
- Compare attributes fields in `bool operator==(const OutputObject& x, const OutputObject& y);` in _output_object.cpp_
- Change the `unordered_set` type to a `vector` type

These 2 solutions give the same result. It works when there is only one zoom level but creates too much features when I tried:

```
{
    "layers": {
        "layer1": { "minzoom": 16, "maxzoom": 18 }
    },
    "settings": {
        "minzoom": 16,
        "maxzoom": 18,
        "basezoom": 18,
        "include_ids": false,
        "name": "OSM Vector Tiles",
        "version": "0.1",
        "description": "Vector Tiles from OSM data, created with tilemaker",
        "compress": "none"
    }
}
```

Have you an idea of How can we fix this problem ?

And thank you for your nice job :)

Thibaud

grafi-tt:
Probably your settings of `maxzoom` and `basezoom` are problematic.
Your value 18 is greater than 16, however, Tilemaker assumes a zoomlevel is less than or equals to 16. (Tilemaker assigns 16bits to each of x and y coordinates.)
If you change the value to 15 or so, I expect the first solution gives the correct result.
Could you try it?

Even though, the second solution would not gives the correct result.
This is because multiple identical objects are copied into `generatedIndex[newIndex]`, in the following part of `tilemaker.cpp`.
```
			// otherwise, we need to run through the z14 list, and assign each way
			// to a tile at our zoom level
			for (auto it = tileIndex.begin(); it!= tileIndex.end(); ++it) {
				uint index = it->first;
				uint tilex = (index >> 16  ) / pow(2, baseZoom-zoom);
[...]

PR #49, opened 2016-07-08
Use vector for keeping output objects, and merge output polygons/linestrings with the identical attributes

grafi-tt:
Polygons with the identical attributes can be merged into one polygon, when outputting a vector tiles.
This PR does this optimization.

To do so, output objects with the same attributes should be made adjacent. So we can't use unordered_set.

I suppose this PR also solves Issue #41.

### DIFF
--- a/src/output_object.cpp
+++ b/src/output_object.cpp
@@ -124,11 +124,44 @@ class OutputObject { public:
 	}
 };
 
-// Hashing function so we can use an unordered_set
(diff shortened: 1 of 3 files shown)


systemed:
These are terrific - thank you!

I'll work on getting a new tilemaker release out.

grafi-tt:
I also thank you, for merging many PRs!

PR #51, opened 2016-07-15
Multi threads support for output

grafi-tt:
Use std::thread and make output parallel. Tilemaker becomes 4x faster on ordinal machines.

By default, the number of core is automatically detected.
The number of threads can be explicitly specified by -threads option.


systemed:
This is absolutely brilliant - thank you! (And apologies for taking so long to review and merge it.)

PR #52, opened 2016-07-15
Fill polygons by tiles

grafi-tt:
`fillCoveredTiles` function implements simple algorithm to fill enclosed tiles, and it is used to correctly output (multi)polygons.
For each x, the range between smallest y and largest y is filled.
If a polygon is concave, this method can yields unnecessary tiles.
But such inaccuracy is harmless.

By this change, relations will check tiles by its own. So tilemaker is changed to 2pass.

`insertIntermediateTiles` function was not correct.
tile-corner
Such error might become more noticeable by filling tiles. So I fixed the function.


systemed:
Another great one, thank you.

I've added a brief check in case constituent ways are missing from the multipolygon (this happens sometimes with Geofabrik extracts, for example).

PR #79, opened 2017-03-30
Use Travis for CI

pnorman:
Replaces #78 

This just does a basic test that everything compiles.


systemed:
Thanks @pnorman and @tomchadwin!

PR #81, opened 2017-05-12
WIP: Tilemaker collects relations which belong to each way

hpgmiskin:
I have made a start at associating relations with ways which can be passed to the Lua `way_function` as a second parameter to improve the creation of vector tiles. This is very much work in progress as currently `tilemaker.cpp` only passes a vector of relation ids as strings to the `way_function`. When I have time I will continue to improve the following:

- Pass relations objects with their associated key value tags
- Move the `WayRelationStore` to the `OSMStore`

When working on this PR I have found the following links useful. Perhaps some could be included in the contributing section? I am happy to open another PR depending on feedback.

 - [`osmformat.proto`](https://github.com/substack/osm-pbf-parser/blob/master/lib/osmformat.proto#L237) the protocol buffer definition for OSM data
 - [`bicycle.lua`](https://github.com/Project-OSRM/osrm-backend/blob/experimental/route_relations/profiles/bicycle.lua#L134) the `osrm-backend` lua profile syntax for using routes passed to the `way_function`

### DIFF
--- a/src/tilemaker.cpp
+++ b/src/tilemaker.cpp
@@ -30,6 +30,7 @@
 typedef unsigned uint;
 #endif
 // Protobuf
+// https://github.com/substack/osm-pbf-parser/blob/master/lib/osmformat.proto
 #include "osmformat.pb.h"
 #include "vector_tile.pb.h"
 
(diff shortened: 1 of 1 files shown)


TimSC:
Can you update process.lua to be in sync with your changes?

systemed:
Closing this PR as stale but I hope we'll be able to pick this functionality up in a future version.

Issue #69, opened 2017-02-13
Tolerate missing node refs in ways


systemed:
Clipped .pbfs produced by Osmosis have a nasty habit of referring to non-existent nodes in ways (see #31, #68).

PR #84, opened 2017-06-23
Add support for 64bit way ids and error handling for missing nodes

TimSC:
Invalid ways are ignored. This addresses #69


systemed:
Terrific - thank you.

Issue #85, opened 2017-06-26
Filling lakes, waterways, etc.


ftheriault:
I am not sure if this is an issue with Tilemaker or if it is only from my understanding of the tool, but I seem to have issues filling river and lakes.

Perhaps if there was documentation on the proper config/process.lua to use when exporting water, the issue wouldn't occur. Anyone willing to share their proper config/process files to be able to correctly fill river and lakes?

The following image shows my rendering.

![water-rendering](https://user-images.githubusercontent.com/619808/27545986-584c861e-5a5f-11e7-95c0-fbcd8a63384f.png)

This image was rendered using OL3 and around the main area, it is supposed to be filled with water. Also, I have used Tilemaker against a GeoFabrik "North America/Canada/Quebec-latest.osm.pbf" file. 

Any help would be greatly appreciated.

TimSC:
Apparently, the second argument in way:Layer("layer_name", is_area) must be true (within process lua file). See CONFIGURATION.md for more info. I'm having difficulty getting areas to appear in my style... we might have the same problem.

ftheriault:
I saw about the is_area parameter, and water was set to true... Perhaps we have the same issue.

 I have tried many ways, but without luck. Here is a section of my current process.lua file. There must be extra data put in the tiles, but I had to try...

`
function way_function(way)
    local waterway = way:Find("waterway") 

    if waterway=="riverbank" or waterway=="river" or
        way:Find("water")=="lake" or
        way:Find("natural")=="coastline" or
        way:Find("natural")=="water"
    then
        way:Layer("water", true)
        way:Attribute("name", way:Find("name"))
        way:Attribute("waterway", waterway)
        way:Attribute("water", way:Find("water"))
        way:Attribute("natural", way:Find("natural"))
    end
`

TimSC:
@ftheriault If you turn on verbose output, do you get anything?

ftheriault:
Actually yes, I have so many lines similar to this :

Exception when writing output object 4294965929 of type 2

Any idea what it could be?

systemed:
Could you share your current process.lua and config.json, so that we can try and reproduce this (with the Quebec extract)?

TimSC:
I was just going to add I had a similar issue which I have since resolved - I'm just trying to remember the exact cause!

ftheriault:
Sure. This is my current configuration.

config.json
```
{
	"layers": {
                "landuse" : { "minzoom" : 8, "maxzoom": 15},
[...]

PR #86, opened 2017-06-27
Fix problematic map data or ignore if we cannot, move worker lambda to its own function

TimSC:
Mainly adding error handling to improve reliability.

I started to a style that is based on the <del>opentilemap's</del> OpenMapTiles' schema, but there is a long way to go. https://openmaptiles.org/schema/

Addressing #85 is a work in progress...


systemed:
This is looking really good!

TimSC:
I think this stuff is ready to merge(?). I'll continue working on the opentilemap style in a new branch.

systemed:
This is great - thanks Tim. Works really well.

PR #112, opened 2018-04-30
Enable high zoom level

ThibaudM:
When the zoom level is equal or greater than 17, the tiles were not generated.
This problem was mainly due to the uint_32 limitation of the tiles indices.

This pull request removes this limitation.

### DIFF
--- a/include/coordinates.h
+++ b/include/coordinates.h
@@ -34,7 +34,7 @@ double tiley2latp(uint y, uint z);
 double tiley2lat(uint y, uint z);
 
 // Get a tile index
-uint32_t latpLon2index(LatpLon ll, uint baseZoom);
+uint64_t latpLon2index(LatpLon ll, uint baseZoom);
 
 // Earth's (mean) radius
 // http://nssdc.gsfc.nasa.gov/planetary/factsheet/earthfact.html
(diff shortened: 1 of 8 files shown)


systemed:
This looks great - thank you. Is there any impact on memory consumption?

ThibaudM:
Yes probably, but I didn't benchmark it. In fact, I'm only working on small datasets, [but at high zoom levels](http://indoor.thibaud-michel.com/caserne/)

TimSC:
@ThibaudM Is there much point in generating the base layer at such a high zoom level? I just wondered what justified such a high accuracy.

ThibaudM:
Yes, you can follow my link in my previous message. I'm working on indoor mapping

systemed:
I think my instinct is to merge a version of this but with a compile-time flag (like the COMPACT_NODES and COMPACT_WAYS flags we already have), so users can choose between the high-zoom tiles and keeping memory consumption down.

TimSC:
Ah indoor mapping, that makes sense now!

pnorman:
I wouldn't oppose a compile-time flag, but I don't think it's the best way to do high accuracy indoor stuff.

A z16 tile at the default 4096 extent is 149mm resolution for tiles at the equator. If you up the extent to 65536 you get 9.3mm resolution. OSM data normally has a resolution of 11mm at the equator.

Going past z16 might be worthwhile for reduced data size but that's a lot more complicated question, because it depends on what area you expect the user to view, how much you can filter out at z16 that you need at z17, etc.

TimSC:
I've merged this branch with my tidy branch https://github.com/systemed/tilemaker/pull/118

TimSC:
This PR can be closed as I've already incorporated it into https://github.com/systemed/tilemaker/pull/119 and it was merged into master. The default is for longer 64-bit tile indices, which are two TileCoordinate objects. Shorter ones can be used by a compile tile flag.

#ifdef COMPACT_TILE_INDEX
typedef uint16_t TileCoordinate;
#else
typedef uint32_t TileCoordinate;
#endif

PR #117, opened 2018-05-18
General code tidy and reorganize

TimSC:
Mainly this splits tilemaker.cpp into more source files and splits the source into several managable functions. This should help further development and get me back up to speed to look at #85 

There should be minimal functional change.


systemed:
Excellent - thank you. Really good to get the monolithic main file refactored like this.

PR #118, opened 2018-05-19
Reorganize SharedData and OSMObject classes

TimSC:
I've tried to make the OSMObject object only available to the data loading phase.

The SharedData object is only used in the multi-threaded output phase.

class Config keeps the config settings together and are read in a more organized way. The shp file loading has been spilt out.

I don't fully understand hasClippingBox and clippingBox in tilemaker.cpp and if the config file or the first pbf should take precidence?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,6 @@ install:
 	install -m 0755 tilemaker /usr/local/bin
 
 clean:
-	rm -f tilemaker src/tilemaker.o
+	rm -f tilemaker src/*.o
 
 .PHONY: install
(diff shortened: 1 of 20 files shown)


TimSC:
@systemed Please merge this when convenient. I think it's as complicated as we might want a PR to get. It seems to work ok in my tests.

I've integrated the ThibaudM:high-zoom-level code into this branch. It is controlled by a compiler flag COMPACT_TILE_INDEX. (It took me quite a while!)

systemed:
Terrific - thanks for this. I'll give it a proper once-over (though your C++ is better than mine! :) ) and merge, probably the start of next week if that's ok.

Issue #115, opened 2018-05-11
Big slowdown from merging geometries


systemed:
When outputting features to the vector tile, currently we merge successive geometries of the same type and attributes: https://github.com/systemed/tilemaker/blob/master/src/tilemaker.cpp#L145

This causes a 100%+ slowdown. Without merging:

    real	0m24.999s
    user	0m28.820s
    sys	0m1.082s

With:

    real	0m53.344s
    user	1m23.208s
    sys	0m4.432s

Filesize is 7.9Mb unmerged, 7.2Mb merged. (Oxfordshire extract, waterways style.)

Ideally we should eliminate the slowdown, but if we can't, this should be a configurable option.

TimSC:
I've added a config and command line option to disable merging geometries in b265e84

PR #119, opened 2018-06-06
Working toward lazy processing and scaling up to larger maps

TimSC:
This PR build on https://github.com/systemed/tilemaker/pull/118 and https://github.com/systemed/tilemaker/pull/112

The highlights:

* PBF reader is more self contained
* The TileIndex has been changed to a vector of std::shared_ptrs, which allows different classes derived from OutputObject
* Command line option (and config option) to disable merging of geometries #115

I am working towards making the processing pipeline support loading map data in a lazy way from any source. This could include loading from PBF tiles, a database, etc. Still quite a bit of work to do though!

Update: new features in later commits:

* The merging of tileIndex data and sorting to lexicographical order is now done in the worker threads and (this is the important part) done on a tile-by-tile basis. (I still need to make the lua processing stage more flexible to really take advantage of this.)


TimSC:
@systemed I am still refactoring stuff. I noticed the Lua functions FindIntersecting and Intersects seems to query shape file layers while Lua is processing the pbf. This intuatively seems hacky but I'm not sure how it's intended to be used. Is it acceptable to just rip out that functionality?

systemed:
It's really important for how I use tilemaker :) - it lets you do things like slightly different rendering for urban and rural areas, for example, by feeding in a set of built-up area polygons.

TimSC:
I have basically finished refactoring the input parsing and data processing functionality. The output functionality could do with some work but it doesn't seem to be urgent. (I've tried to retain all previous functionality in Lua but not well tested.)

I think this branch is ready for testing and eventual merging. @systemed

systemed:
This looks terrific - thank you. I'll test it over the next couple of days.

systemed:
I've been playing around with this. Running it with the default configuration on oxfordshire-latest.osm.pbf (from Geofabrik) always reports:

> libc++abi.dylib: terminating with uncaught exception of type kaguya::LuaRuntimeError: Could not find node 140591891834448

The node changes each time but is of a similar magnitude: 140597775794752, 140638078329280, 140188761944144, 140541278298896.

Any thoughts?

TimSC:
I'll take a look when I come back from camping.

TimSC:
@systemed The missing node bug is resolved in 4570ab2

ThibaudM:
Thank your for this great work @TimSC !

I'm encountering a problem with your last version. The first image is what I obtain with `master` and the second image is what I obtain with this PR.
screenshot_20180724-145754 screenshot_20180724-145802

I was trying to figure it out and I had a look at the generated geojson.

From this "simple" quadrilateral:
```
<osm version='0.6'>
  <node id='-45213' lat='45.21785043007' lon='5.80703391154' />
  <node id='-45214' lat='45.21786560012' lon='5.80706612458' />
  <node id='-45215' lat='45.21782063055' lon='5.80710678249' />
  <node id='-45216' lat='45.2178054467' lon='5.80707442968' />
  <way id='-45212'>
    <nd ref='-45213' />
    <nd ref='-45214' />
    <nd ref='-45215' />
    <nd ref='-45216' />
    <nd ref='-45213' />
  </way>
</osm>
```
I should obtain two tiles (because the polygon should be included in both):
**18/135300/94074.geojson**
```
{
    "geometry": {
        "coordinates": [
            [
                [2361, 106],
                [2441, -20],
                [2304, -20],
                [2265, 42],
                [2361, 106]
            ]
        ],
        "type": "Polygon"
    },
    [...]
}
```
and **18/135300/94075.geojson**
```
{
    "geometry": {
        "coordinates": [
            [
                [2482, 4012],
                [2386, 3947],
                [2278, 4116],
                [2415, 4116],
                [2482, 4012]
            ]
        ],
        "type": "Polygon"
    },
    [...]
}
```

But, from this PR, I obtain two different tiles:
**18/135300/94074.geojson**
```
{
    "geometry": {
        "coordinates": [
            [
                [2265, 42],
                [2361, 106],
                [2441, -20]
            ], [
                [2304, -20],
                [2265, 42]
            ]
        ],
        "type": "MultiLineString"
    },
    [...]
}
```
and **18/135300/94075.geojson**
```
{
    "geometry": {
        "coordinates": [
            [2415, 4116],
            [2482, 4012],
            [2386, 3947],
            [2278, 4116]
        ],
        "type": "LineString"
    },
    [...]
}
```
Both are not well generated. This happens only for features which cross two tiles.
I hope this will help

TimSC:
@ThibaudM Thanks for testing! I've starting trying to identify the cause of the bad polygon shapes.

dd33013a07f7131ecaecb5a5e11b414b370a414a doesn't have the problem apparently.
2e612b88aa5aea5a884391e0e77d2d713ac8d730 has the bug

Watch this space...

TimSC:
@ThibaudM I found the problem: ways were being treated as Linestrings and clipped using boost::geometry rather than clipper.

systemed:
Looking good - thanks! I ran it on an Oxfordshire extract again earlier today and it completed fine without errors.

Currently trying it on england-latest.osm.pbf and it appears to have stalled somewhere between writing out z9 and z10, but that may be the 'combine objects' code again - will try again without it. (But not until next week as I'm just setting off for SOTM!)

systemed:
Just completed:

    real	292m57.817s
    user	719m41.045s
    sys	7m59.581s

I'll have a go with combine objects disabled.

Tiles generally looking really good. Some polygon filling issues:

screen shot 2018-07-25 at 22 37 38

This is on the NW side of Oxford and appears to change at different zoom levels, so I'm presuming it's a clipping issue. Will investigate further when I'm back.

ThibaudM:
It works for me now, good job!
And maybe see you at SOTM.

systemed:
No joy at fixing the clipping issue, I'm afraid. It's evidently something in `ClipGeometryVisitor::operator()(const MultiPolygon &mp)`, but whereas previously I'd been able to get round this by just using `geom::intersection`, that doesn't help here. I'll keep niggling away at it but would be delighted if someone beat me to it. :)

TimSC:
@systemed Where abouts are you seeing problems? Do you also have the same problems with dd33013 ?

systemed:
@TimSC Just tried with that revision and it doesn't appear to write anything at all from the .pbf to the .mbtiles, I'm afraid! (Though it does write data from shapefiles...)

Currently `master` exhibits the same behaviour (i.e. leaking multipolygons). Patching it so that the Clipper-based `ClipGeometryVisitor:operator()(const MultiPolygon &mp)` is replaced with a simple call to Boost's `geom::intersection` works:

````
Geometry ClipGeometryVisitor::operator()(const MultiPolygon &mp) const {
	Polygon clippingPolygon;
	geom::convert(clippingBox,clippingPolygon);
	if (geom::within(mp, clippingPolygon)) { return mp; }

	MultiPolygon out;
	geom::intersection(mp,clippingBox,out);
	return out;
````

When I tried the same patch on this PR it didn't seem to work, although it might be worth retrying.

TimSC:
@ThibaudM @systemed I found a bug in ConvertFromClipper which caused inner polygons to be wrong or missing. Plz give the latest version a go! (16fe95c or later)

systemed:
Looking good - am on holiday atm but will check out when I’m home.

systemed:
I think you've cracked it - all the polygons now appear to be well-formed at all zoom levels. \o/

TimSC:
@systemed Great stuff! Ready for merge? 

I have another batch of changes but they will need to rebased (eventually).

systemed:
Merged! Thanks again.

PR #128, opened 2019-02-01
Reading input from tiled pbf files

TimSC:
Input data can be loaded from tiled pbf files. The input tiles are in the usual slippy tile naming convensions. Tilemaker needs the folder containing the tiled data as a program argument:

    tilemaker /path/to/tiledata --tiled-input --config resources/config-openmaptiles.json --process resources/process-openmaptiles.lua --output=tiles/

The advantage of using a tiled pbf input is that the memory usage is much lower, even for large areas. For a small area, the file structure of the tiled data is as follows:

    12/2030/1377.pbf
    12/2030/1378.pbf
    12/2031/1377.pbf
    12/2031/1378.pbf

Example tiled data can be downloaded from: https://archive.org/details/isle-of-wight-fosm.tar

Currently, shapefiles are loaded into memory, so this is still a limiting factor in processing large areas.


systemed:
Wow. Neat stuff. Do you have a suggested pipeline for creating .pbf tiles?

TimSC:
On generating tiled pbf files, here is what comes to mind:
* Repeatedly calling the server API to get the tiles. On the public server, this quickly hits request limits so it really has to be done on a private server with the relevant data.
* Distributing pre-generated tiled data would be a good idea. Looks like people have been playing with the idea for some time https://wiki.openstreetmap.org/wiki/Tile_data_server
* Using osmosis or osmconvert to repeatedly crop data tiles out of a medium sized dump. This is probably an inefficient but easy to implement approach.
* I experimented with splitting a planet dump into a quad tree. This can be recursively done at different levels to produce zoom 12 tiles (or whatever we want). It was a slow process and probably not much different from using a private map server.
* We could add a preprocessing stage to tilemaker in which a pbf dump is read into a sqlite database. This is then queried to produce data tiles on the fly. This approach is used by OsmAndMapCreator. The trouble is this only scales up to a medium sized country. I tried importing a planet dump into sqlite and it eventually grinds to a halt.

Oh this PR doesn't solve the problem of generating low zoom map tiles either. They still require a massive amount of RAM.

systemed:
Just experimenting with this and I think there's a regression with shapefiles. tilemaker is reading the shapefiles in happily (and complaining about invalid entities now and then) but they're not written to the .mbtiles - they don't show up on a rendered map, and the output file is smaller (to the extent that I'd expect if shapefile data wasn't written).

Config lines are things like

    "dlua"    : { "minzoom": 9, "maxzoom": 14, "source": "/Users/richard/Maps/shapefiles/dlua_wgs84.shp", "simplify_below": 13, "simplify_level": 0.0005, "index": true, "index_column": "NAME" },
    "canals"  : { "minzoom": 7 , "maxzoom": 14, "source": "Data/Output/ct_waterways.shp", "simplify_below": 13, "simplify_level": 0.0005, "source_columns": ["ct_id","name"] },
    "towpath" : { "minzoom": 12, "maxzoom": 14, "source": "Data/Output/ct_towpaths.shp" },

Any thoughts?

systemed:
Incidentally, this might be interesting for tiled .pbf files:

* https://www.openstreetmap.org/user/SimonPoole/diary/47275
* https://github.com/simonpoole/mapsplit/tree/v2

Obviously it still requires plenty of RAM to do the splitting, but potentially the .mbtiles could be hosted somewhere and made available for download.

TimSC:
My first guess is that these are linear features. I've not tested that at all. I'm thinking we need to start on some demo data and unit tests. Struggling to write value tilemaker rules as I'm a bit rusty. @systemed Basically can you lend a hand in reproducing this, with an updated style and a few shapefiles?

systemed:
They are linear features, yep. I should be able to find a test case.

TimSC:
I found the problem: the shapefile clipping was in a different projection than I expected. Fixed in 
d040ffa 

I also added a massive speed up to shapefile reading in f4d97bc !

@systemed can you test your custom styles with this branch plz?

TimSC:
I was looking at merging changes from master into this branch. Commit 06ccc8d043410b3f50dd241f03f04e001ae153e0 is going to be a total nightmare, since it changes the relationships between the class and I don't have a working example of the lua configuration. I've done major refactoring in my scaleup3 and scaleup4 branches on shapefile reading. I think the best move is to first check what I have in scaleup4 is working (with some decent test data of Intersect and FindIntersecting), I get a working configuration for the remapped shapefile lua and I reimplement 06ccc8d043410b3f50dd241f03f04e001ae153e0 all over again at the tip of my scaleup4 branch. Any thoughts @systemed?

systemed:
Quick note to say I've been up to my eyeballs recently but I'm hoping to get to this after Easter.

systemed:
I should probably have said Easter 2020. ;) Supplanted by https://github.com/systemed/tilemaker/tree/mapsplit which is inspired by this PR.

PR #130, opened 2019-03-08
WIP: Add test data

TimSC:
I've added some test data. I've used the Isle of Wight, UK as the test area. This is to allow for people to get the tool running with the minimum of hassle and hopefully to allow unit tests to be written.

The data is from https://data.cdrc.ac.uk/dataset/cdrc-2015-os-geodata-pack-isle-of-wight-e06000046 with is UK Open Government Licence (OGL). It originally contained shape files in OS Grid projection. I've converted some to OSM format and others to WGS 84 shapefiles.

Unit tests probably should use a smaller area because it takes a while to process even this amount of data.

I also noticed there are problems in low zoom rendering of sea polygons, which I will investigate.

This PR is not ready to merge yet! Discussion needed as to the best approach.


TimSC:
@systemed I'm having a bit of trouble writing the config.json to pull in the woodland shapefile. Any idea where I am going wrong? https://github.com/TimSC/tilemaker/blob/testdata/resources/config-testdata.json

TimSC:
@systemed It would be really helpful to have a semi-realistic example of using the lua FindIntersecting and Intersects functions. I've been refactoring them on a different branch but I don't have any way to test them properly.

systemed:
Closing this PR as stale but I'll look through and see what looks relevant for applying to the current codebase.

PR #141, opened 2019-09-18
Add missing CPP files to CMakeLists.txt for tilemaker

ogre:
CMake configuation was missing many CPP files to build/link tilemaker, at least under ubuntu 18.04.


systemed:
Thank you!

PR #148, opened 2020-02-24
Improve configuration for openmaptiles

typebrook:
Make layers of water* more consistent with openmaptiles v3.11

Of course it is not one-to-one mapping, openmaptiles uses area to filter
out small water-body in lower zoom levels, but we cannot.

@systemed Is it possible to introduce `way:Area` into Lua spatial queries? I don't write C++, but maybe there are some stacks can be applied to tilemaker inside [mapbox/earcut.hpp](https://github.com/mapbox/earcut.hpp)

### waterway
- Removed from z8
- Only include waterway=river with name in z9-z11
  Also, do simplification below z12

### water
- Removed from z8
- Do simplification below z12

### water_name
- Removed from z8
- Put waterway=dock, natural=water, natural=bay or landuse=reservoir
  into z12 and above(originally only appears in z14)

---
Comparison with [mbview](https://github.com/mapbox/mbview) in z9

openmaptiles
omt

before
before

after
Screenshot from 2020-02-24 14-21-23

### DIFF
--- a/resources/config-openmaptiles.json
+++ b/resources/config-openmaptiles.json
@@ -8,7 +8,8 @@
 
 		"housenumber":      { "minzoom": 14, "maxzoom": 14 },
 
-		"waterway":         { "minzoom": 8,  "maxzoom": 14 },
+		"waterway":         { "minzoom":  9,  "maxzoom": 14, "simpolify_below": 12, "simplify_length": 2, "simplify_ratio": 2},
+		"waterway_detail":  { "minzoom": 12,  "maxzoom": 14, "write_to": "waterway" },
 
 		"transportation":             { "minzoom": 8,  "maxzoom": 14, "simplify_below": 13, "simplify_level": 0.0003 },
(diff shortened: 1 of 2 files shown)


systemed:
Thank you!

Not 100% sure what you mean about `way:Area` - could you give an example of what you'd like to do?

typebrook:
> Not 100% sure what you mean about way:Area - could you give an example of what you'd like to do?

get the area of a closed way(polygon) with square meters, then I can filter out small polygons in Lua script. According to [`CONFIGURATION.md`](https://github.com/systemed/tilemaker/blob/master/CONFIGURATION.md), Lua spatial queries currently only supports `Intersects` feature.

systemed:
Ah! way:Area() should already be implemented but evidently not documented - try it and see.

typebrook:
Wow! my bad, I should check source file at first :sweat_smile: 
https://github.com/systemed/tilemaker/blob/153d35d8d69790253a9521a2956483df1cc6e890/src/osm_lua_processing.cpp#L152-L161
That's great!Then I'll keep improving openmaptiles schema implementaion next week. Thanks for your reply!

systemed:
My fault for not putting it in the docs! Thanks for your ongoing work on this - really pleased to see it.

PR #151, opened 2020-04-28
Update config for openmaptiles

typebrook:
Add aerodrome_label layer, see [OpenMapTIles schema](https://openmaptiles.org/schema/#aerodrome_label)

Result:
[TSA](https://en.wikipedia.org/wiki/Songshan_Airport) in Taiwan
Screenshot from 2020-04-28 22-25-46


systemed:
Looks great - thank you!

PR #152, opened 2020-05-05
Use Spherical Strategy to calculate area

typebrook:
@systemed 
Looks like the original area calculation is meaningless with 2D points for tiles.

Here I use [for_each_point](https://www.boost.org/doc/libs/1_72_0/libs/geometry/doc/html/geometry/reference/algorithms/for_each/for_each_point.html) to get original latitude value and [area function](https://www.boost.org/doc/libs/1_72_0/libs/geometry/doc/html/geometry/reference/algorithms/area/area_2_with_strategy.html) with strategy of spherical earth. So the result is in square meters.

I am new to C++ and not familiar with its conventions. So if anything looks weird, please don't hesitate to ask me to fix.

--
Also, some problem confuse me: When I tried to use geographic strategy with WGS84 spheroid, the result is more inconsistent with the area calculate by `GDAL` tools. So here I picked strategy:
```C++
boost::geometry::strategy::area::spherical<> sph_strategy(6371008.8);
```
rather than
```C++
boost::geometry::srs::spheroid<double> spheroid(6378137.0, 6356752.3142451793);
boost::geometry::strategy::area::geographic<> geo_strategy(spheroid);
```
You may check the example of strategies above in [area function](https://www.boost.org/doc/libs/1_72_0/libs/geometry/doc/html/geometry/reference/algorithms/area/area_2_with_strategy.html)

### DIFF
--- a/include/geomtypes.h
+++ b/include/geomtypes.h
@@ -12,6 +12,7 @@
 #include <boost/geometry/geometries/point_xy.hpp>
 #include <boost/geometry/index/rtree.hpp>
 typedef boost::geometry::model::d2::point_xy<double> Point; 
+typedef boost::geometry::model::point<double, 2, boost::geometry::cs::spherical_equatorial<boost::geometry::degree> > DegPoint;
 typedef boost::geometry::model::linestring<Point> Linestring;
 typedef boost::geometry::model::polygon<Point> Polygon;
 typedef boost::geometry::model::multi_polygon<Polygon> MultiPolygon;
--- a/src/osm_lua_processing.cpp
+++ b/src/osm_lua_processing.cpp
@@ -149,14 +149,31 @@ double OsmLuaProcessing::ScaleToKiloMeter() {
 	return (1/1000.0) * ScaleToMeter();
 }
 
+void reverse_project(DegPoint& p) {
+    geom::set<1>(p, latp2lat(geom::get<1>(p)));
+}
+
 // Returns area
 double OsmLuaProcessing::Area() {
 	if (!IsClosed()) return 0;
+
+	std::ostringstream stream;
+	geom::strategy::area::spherical<> sph_strategy(6371008.8);
+
 	if (isRelation) {
-		return geom::area(multiPolygonCached());
+		stream << setprecision(10) << geom::wkt(multiPolygonCached());
+		geom::model::multi_polygon<geom::model::polygon<DegPoint> > geo_poly;
+		geom::read_wkt(stream.str(), geo_poly);
+		geom::for_each_point(geo_poly, reverse_project);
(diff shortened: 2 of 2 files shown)


systemed:
Thank you (and sorry for the delay in reviewing). This looks great.

systemed:
Spoke too soon! This builds fine on my machine (Boost 1.69) but is failing on the automated build (1.65).

systemed:
Fixed. I've reverted to the old code for pre-1.67 while still using the new code for 1.67 upwards.

Streaming everything into a .wkt and then re-parsing it seems inefficient - can we skip that somehow?

typebrook:
> Streaming everything into a .wkt and then re-parsing it seems inefficient - can we skip that somehow?

I am afraid not. Tried using [`transform`](https://www.boost.org/doc/libs/1_72_0/libs/geometry/doc/html/geometry/reference/algorithms/transform/transform_2.html) function to convert polygon with `Point` to polygon with `DegPoint`, but code like this [gist](https://gist.github.com/7db687c00fd2d45cc3f882afeefb667a) just get compiler error.

Also, it is not OK to totally replace `Point`(2D) with `DegPoint`(3D). Because [`centroid`](https://www.boost.org/doc/libs/1_72_0/libs/geometry/doc/html/geometry/reference/algorithms/centroid/centroid_3_with_strategy.html) function does not support 3D polygon now.

So streaming 2D polygon to wkt and read it into 3D polygon is the only solution AFAIK. If you come out a better way, then just make another patch.

typebrook:
And `OsmLuaProcessing::Length()` is also need to calculate by 3D strategy, I'll make another PR later.

systemed:
Ok, I've fixed this to avoid the wkt step in https://github.com/systemed/tilemaker/commit/5ea32b8bea327140068acac5c571621ba1fc46b1

Polygons are easy: we can just dupe the polygon with `geom::assign`, unproject it back to lat/lon, and then use `geom::area` with the spherical strategy. Unfortunately that's not implemented for multipolygons. However, we can get round that by calculating the area of each member polygon, then summing them.

typebrook:
Wow! How could I miss `geom::assign` before.

> we can get round that by calculating the area of each member polygon, then summing them.

So if a multipolygon is hollow, then its child polygons would have opposite winding order like image below. And summary area with positive and negative value, right?

Screenshot from 2020-05-19 20-56-06

systemed:
Boost multipolygons are a bit different to OSM ones. In Boost, [a normal "polygon" can have holes (inner rings)](https://www.boost.org/doc/libs/1_65_1/libs/geometry/doc/html/geometry/reference/models/model_polygon.html). A Boost "multipolygon" is a collection of one or more polygons.

So `geom::area` on the member polygons should already take account of the holes.

PR #155, opened 2020-05-11
Use Ubuntu 20.04 as the base image of the Dockerfile

leonardehrenfried:
This upgrades the base image to Ubuntu 20.04 LTS.

This of course means that the packages like `libboost-all-dev` also have been upgraded. Would it be problematic to use newer versions?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
-FROM ubuntu:16.04
+FROM ubuntu:20.04
 LABEL Description="Tilemaker" Version="1.4.0"
+
+ARG DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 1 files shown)


systemed:
Thanks! The Dockerfile doesn't have anyone actively maintaining it so I'm happy to merge this on sight.

leonardehrenfried:
Thanks!

I'm quite happy to become the go to Docker guy and fixer-upper of the file.

I just tested it and I could successfully build mbtiles for Liechtenstein. In the next couple of days I will build larger areas and will make more PRs if I notice problems.

systemed:
Excellent - thanks!

PR #156, opened 2020-05-13
Reduce the size of the docker image

leonardehrenfried:
This PR reduces the number of dependencies installed when building the docker images to the minimum number required. It also uninstalls several of them after building tilemaker.

I also switched to the slim Debian base image.

These two measures reduce the size of the image from 1GB to 350MB.

You could potentially uninstall more packages but to me that seemed like a good enough improvement for now.


systemed:
Wow, that's definitely a worthwhile improvement. Thanks!

PR #161, opened 2020-05-21
Use Spherical Strategy to calculate length

typebrook:
continue #152

@systemed 
Successfully build on my machine with Boost 1.72.
Not sure it would be fail on automated build (Boost 1.65), please take a look.

Also, `boost.geometry.length` would calculate all multi_polygon as zero, so just return 0 if feature is a relation.

ref:
https://www.boost.org/doc/libs/1_72_0/libs/geometry/doc/html/geometry/reference/algorithms/length/length_2_with_strategy.html#geometry.reference.algorithms.length.length_2_with_strategy.behavior

### DIFF
--- a/src/osm_lua_processing.cpp
+++ b/src/osm_lua_processing.cpp
@@ -191,11 +191,13 @@ double OsmLuaProcessing::Area() {
 
 // Returns length
 double OsmLuaProcessing::Length() {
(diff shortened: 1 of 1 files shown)


systemed:
Fine all the way back to 1.57 at least. Thanks!

PR #164, opened 2020-05-24
Update config for OpenMapTiles

typebrook:
Continue #148. Time to apply `way:Area()` in Lua script!

b0992cb
- Add more tags into waterway layer and water layer with OpenMapTiles spec.
- Explicitly only include linestrings into waterway, and polygons into water. This is because of some OSM features are not well tagged (For example: `waterway=riverbank` on a non-polygon way, see the case in Taipei below)
Screenshot from 2020-05-25 01-34-32

6fe2cfc
- Filter out small water body in lower zoom levels, this should dramatically reduce tile size in lower zoom level

### DIFF
--- a/resources/config-openmaptiles.json
+++ b/resources/config-openmaptiles.json
@@ -21,10 +21,16 @@
 
 		"building":          { "minzoom": 13, "maxzoom": 14 },
 
-		"water":             { "minzoom": 9,  "maxzoom": 14, "simplify_below": 12, "simplify_length": 1, "simplify_ratio": 2},
+		"water":             { "minzoom": 12, "maxzoom": 14, "simplify_below": 12, "simplify_length": 1, "simplify_ratio": 2},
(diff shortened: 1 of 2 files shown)


systemed:
Superb. Thanks!

I'm _considering_ having an option to set the minimum zoom level for an individual feature from the Lua script, so that you could say `way:SetMinZoom(12)` and it would only appear at z12+ upwards. That would relieve the need to define so many additional layers.

typebrook:
> you could say way:SetMinZoom(12) and it would only appear at z12+ upwards

This sounds great!

systemed:
Now implemented in https://github.com/systemed/tilemaker/commit/080f94d4715ba2000388a52db6c66e7fd67f7a11

It's pretty simple - just call `way:MinZoom(12)` and the way won't appear at below z12. Note that the JSON layer minZoom is still observed.

Best bit is that this doesn't require any extra memory - it's stashed in four spare bits of OutputObject. There's a lot more we can do on memory saving but more of that anon!

typebrook:
@systemed 
This is amazing! So productive man!

ogre:
Thank you for that !
Just when I needed it !

PR #165, opened 2020-05-26
Update build config and README.md from c++11 to c++14

ogre:
g++ needs C++14 configuration to successfully build latest tilemaker.
Otherwise it fails with internal compiler error (boost 1.68) or other boost errors (boost 1.73).


systemed:
Done (actually in #229) - thanks!

PR #166, opened 2020-05-27
Improve GitHub Action

typebrook:
tilemaker is within Dockerfile, so for now it already could be used as Github Action.
This PR does the following things:
- Add `action.yml` to make the inputs more clear
- Add example of usage into README
- Refactor `ci.yml` with Github Action. So each time we change build steps or dependencies, just need to modify Dockerfile

The CI works same as how @leonardehrenfried  did in #158 , but takes more time to build(1min->2.5min). I guess it is because Github Action calls `docker` to make a new image and create a new container. And what `ci.yml` works now is just build `tilemaker` directly in the workflow runner. Anyway just check the log of CI below.

@systemed 
Also if this PR is merged, since `action.yml` is there, you'll see the hint at repo page:
Screenshot_2020-05-27 OsmHackTW tilemaker
It's up to you to make a new release to promote tilemaker on the marketplace (Or not).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,8 +9,32 @@ env:
   AREA: liechtenstein
 
 jobs:
-  generate-tiles:
+  build:
+    name: Compile, install and build mbtiles
     runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y --no-install-recommends build-essential liblua5.1-0 liblua5.1-0-dev libprotobuf-dev libsqlite3-dev protobuf-compiler shapelib libshp-dev libboost-program-options-dev libboost-filesystem-dev libboost-system-dev
+
+    - name: Build and install
+      run: |
+        make -j 2
+        sudo make install
+
+    - name: Build openmaptiles-compatible mbtiles files of Liechtenstein
+      run: |
+        curl http://download.geofabrik.de/europe/${AREA}-latest.osm.pbf -o ${AREA}.osm.pbf
+        tilemaker ${AREA}.osm.pbf --config=resources/config-openmaptiles.json --process=resources/process-openmaptiles.lua --output=${AREA}.mbtiles --verbose
+
+  Github-Action:
+    name: Generate mbtiles with Github Action
+    runs-on: ubuntu-latest
+
     steps:
     - name: Check out repository
       uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,26 +5,21 @@ on:
   pull_request:
     branches: [ master ]
 
+env:
+  AREA: liechtenstein
+
 jobs:
-  build:
-
-    name: Compile, install and build mbtiles
+  generate-tiles:
     runs-on: ubuntu-latest
-
     steps:
-    - uses: actions/checkout@v2
+    - name: Check out repository
+      uses: actions/checkout@v2
 
-    - name: Install dependencies
-      run: |
-        sudo apt-get update
-        sudo apt-get install -y --no-install-recommends build-essential liblua5.1-0 liblua5.1-0-dev libprotobuf-dev libsqlite3-dev protobuf-compiler shapelib libshp-dev libboost-program-options-dev libboost-filesystem-dev libboost-system-dev
+    - name: Download PBF file
+      run: curl http://download.geofabrik.de/europe/${AREA}-latest.osm.pbf -o ${AREA}.osm.pbf
 
-    - name: Build and install
-      run: |
-        make -j 2
-        sudo make install
-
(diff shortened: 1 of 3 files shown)


typebrook:
```yaml
- uses: systemed/tilemaker@master
```

In README, the example uses `master` branch as reference for Github Action.
This is because in the latest release there is no `action.yml` there, so cannot replace it like `systemed/tilemaker@v1.6.0. This should be fixed whenever a new release is make

systemed:
This looks good to me but I confess I know very little about Docker! @leonardehrenfried, do you have any thoughts on this?

leonardehrenfried:
Personally, I would not use GH Actions to build my tiles since my input is huge, but I like the fact that it not only builds the docker image but also uses it to build an actual mbtiles file. However, I'm not sure if I would completely replace the previous CI step which uses a plain Ubuntu VM to install the packages and build the output. GH Actions makes it easy to do both and that's what I would do.

I have other review comments but I will leave them in the code directly.

.github/workflows/ci.yml
```diff
@@ -5,26 +5,21 @@ on:
   pull_request:
     branches: [ master ]
 
-jobs:
-  build:
+env:
+  AREA: liechtenstein
```
leonardehrenfried, line 9: If you really want to promote tilemaker usage through GH Actions I would specify a complete (and overridable) download URL since right now you would be limited to European countries and you could not use regional extracts or data sources other than Geofabrik.

.github/workflows/ci.yml
```diff
@@ -5,26 +5,21 @@ on:
...
+      run: curl http://download.geofabrik.de/europe/${AREA}-latest.osm.pbf -o ${AREA}.osm.pbf
 
-    - name: Build openmaptiles-compatible mbtiles files of Liechtenstein
-      run: |
-        curl http://download.geofabrik.de/europe/liechtenstein-latest.osm.pbf -o liechtenstein.osm.pbf
-        tilemaker liechtenstein.osm.pbf --config=resources/config-openmaptiles.json --process=resources/process-openmaptiles.lua --output=liechtenstein.mbtiles --verbose
+    - name: Build openmaptiles-compatible mbtiles files of given area
+      uses: ./
+      with:
+        input: ${{ env.AREA }}.osm.pbf
```
leonardehrenfried, line 24: I would not have a problem with these files always having the same name but if you want get fancy you can make them overridable.

.github/workflows/ci.yml
typebrook: I thinks this doesn't matter. What `ci.yml` does is just test the current build for `tilemaker`. 
How user apply GH Action is coded inside `action.yml`

.github/workflows/ci.yml
typebrook: Same. Take a look in `action.yml` and `README.md`

.github/workflows/ci.yml
leonardehrenfried: Ah I see, the download is the responsibility of the "using" action. Then this is a moot point. Ignore me.:)

typebrook:
@leonardehrenfried 
So do you agree to refactor `ci.yml` like this way? At least in the future when build steps changes, we don't need to modify `Dockerfile` and CI flow at the same time.

I also refine the example of GH Action, so this is how user use it:
```yaml
- uses: systemed/tilemaker@RELEASE_VERSION
  with:
    # Required, same to --input
    input: /path/to/osm.pbf
    # Required, same to --output. Could be a directory or a .mbtiles files
    output: /path/to/output
    # Optional, same to --config
    # If not being set, default to resources/config-openmaptiles.config
    config: /path/to/config
    # Optional, same to --process
    # If not being set, default to resources/process-openmaptiles.lua
    process: /path/to/lua
    # Optional, other arguments
    # If not being set, default to '--verbose'
    extra: --threads 0
```

leonardehrenfried:
Well, I would also add a job that uses a plain Ubuntu VM to compile tilemaker and build tiles. So instead of replacing the current job you add your new Docker-based one.

typebrook:
> Well, I would also add a job that uses a plain Ubuntu VM to compile tilemaker and build tiles. So instead of replacing the current job you add your new Docker-based one.

> So you want to make sure build success on Ubuntu, make sense. I'll resume it later, will get back to you when it is done

typebrook:
@leonardehrenfried 
Done, original build job is resumed. Just apply env variable for osm.pbf from Geofabrik.

leonardehrenfried:
Ok, looks good to me then.

systemed:
Thank you!

typebrook:
@systemed 
Thanks!

Then if tilemaker has a new release later, please also update the reference in [`README.md`](https://github.com/systemed/tilemaker/blob/1e8ec731461725dab72dfb5b138f58f1fe409920/README.md#github-action) (from `@master` to `@v1.7.0`).

Also, if you have no idea how to deal with marketplace. We may discuss in a new issue.
Screenshot_2020-05-27 OsmHackTW tilemaker

PR #172, opened 2020-06-22
Turn Lua variables into Sets

leonardehrenfried:
When I tried out the recently updated waterways I noticed that rivers and streams didn't show up on the map.

After much head-scratching I traced it down to the waterway variables being plain Lua hashtables (lists).

Apparently, you need to use the `Set` constructor otherwise the `set[key]` lookup does not work. This make sense as all the other value lists use it.

https://stackoverflow.com/questions/33510736/check-if-array-contains-specific-value


systemed:
Of course - I should have spotted that before. Thanks.

systemed:
See also https://github.com/systemed/tilemaker/commit/2aacdd03acd4f20f08896e92d08d7ac95eafca94 .

PR #175, opened 2020-07-05
Refine openmaptiles config with mountain_peak layer

typebrook:
@systemed 

e64a158
Set all features in mountain_peak as rank=1

Nearly all the styles from klokan(openmaptiles-topo, maptiler-terrain-gl-style, maptiler-topographique) only render mountain_peak feature with rank=1. Which means, if we set them as 5, none of them would not be rendered.

Though setting all features with rank=1 makes some non-important peaks also got rendered on those styles, it is better than nothing. The result would be like this:

**style [maptiles-topographique](https://www.maptiler.com/maps/#topographique//vector/13.78/121.61565/25.01444), left uses tiles from this PR, right uses openmaptiles**
Screenshot from 2020-07-05 22-12-51

e2fac1f
- Add `class`, which is necessary in some style like [openmaptiles-topo](https://www.maptiler.com/maps/#topo//vector/14.13/121.54814/25.09374)
- Include natural=volcano into mountain_peak, see the result with [openmaptiles-topo](https://www.maptiler.com/maps/#topo//vector/14.13/121.54814/25.09374)(The red triangles are volcano outputs):
Screenshot from 2020-07-05 22-04-58

- Floor elevation as Integer, or some terrible thing would happen...like this:
Screenshot_2020-07-05 Mapbox GL Inspect

refs:
- What is rank in layer mountain_peak? How to calculate it when generating tiles?
  https://github.com/openmaptiles/openmaptiles/issues/687

- How a style render mountain_peak noramlly
  https://github.com/openmaptiles/maptiler-terrain-gl-style/blob/6c480e82af32f4df2ce49d85f59e625ef94337f4/style.json#L715

### DIFF
--- a/resources/process-openmaptiles.lua
+++ b/resources/process-openmaptiles.lua
@@ -133,10 +133,11 @@ function node_function(node)
 
 	-- Write 'mountain_peak' and 'water_name'
 	local natural = node:Find("natural")
-	if natural == "peak" then
+	if natural == "peak" or natural == "volcano" then
 		node:Layer("mountain_peak", false)
 		SetEleAttributes(node)
 		node:AttributeNumeric("rank", 1)
+		node:Attribute("class", natural)
 		SetNameAttributes(node)
 		return
 	end
@@ -458,7 +459,7 @@ end
 function SetEleAttributes(obj)
     local ele = obj:Find("ele")
 	if ele ~= "" then
-		local meter = tonumber(ele) or 0
+		local meter = math.floor(ele) or 0
 		local feet = math.floor(meter * 3.2808399)
 		obj:AttributeNumeric("ele", meter)
 		obj:AttributeNumeric("ele_ft", feet)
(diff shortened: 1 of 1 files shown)


resources/process-openmaptiles.lua
```diff
@@ -458,7 +459,7 @@ end
 function SetEleAttributes(obj)
     local ele = obj:Find("ele")
 	if ele ~= "" then
-		local meter = tonumber(ele) or 0
+		local meter = math.floor(ele) or 0
```
systemed, line 462: I think we probably need to do `math.floor(tonumber(ele) or 0)` here - I've been bitten in the past when there are non-numeric values in the `ele` field:

https://taginfo.openstreetmap.org/tags/ele=%3E160

    > print(math.floor(">160") or 0)
    stdin:1: bad argument #1 to 'floor' (number expected, got string)
    > print(math.floor(tonumber(">160") or 0))
    0

typebrook: Make sense! Also some contributor would tag `ele` like: `ele:160m`.

systemed:
Thank you!

typebrook:
Thanks!
@systemed 
Could you also give some feedback on #166 ? I think it also helps on tilemaker promotion

PR #177, opened 2020-07-11
Add water names at lower zoom level

leonardehrenfried:
Since the default maxZoom is 14 the water names would not show up due to them being configured for 16-17. This PR corrects this.

I think it would probably smart to add the names also depending on the size of the body of water, wouldn't it?


systemed:
Good spot. Yes, doing it on the area seems like a good idea too.

PR #179, opened 2020-09-04
Fix up labels for lakes and rivers

leonardehrenfried:
This PR improves the handling of labels for bodies of water.

## Lakes
I noticed on our map that sometimes the labels would show up earlier than the actual lakes, which has been fixed. Also, I've restricted the display of names to actual lakes and excluded things like basins and swimming pools as their nondescript names often are a distraction.

## Rivers

So far it appears that river names were not in the correct format, so I added them.

### Before 
Screenshot from 2020-09-04 14-08-04

### Before 
Screenshot from 2020-09-04 14-02-24

~~However, this part probably needs a little careful review, as i'm not entirely sure why it works. After having changed the layer from `waterway` to `water_name_detail` I expected the rivers themselves to disappear but they are still there.~~

~~Can you help me understand?~~

I think I figured it out. It needed a second pass so that the layers for the water polygon and the name are created separately.

### DIFF
--- a/resources/process-openmaptiles.lua
+++ b/resources/process-openmaptiles.lua
@@ -346,7 +346,7 @@ function way_function(way)
 	-- Set 'waterway' and associated
 	if waterwayClasses[waterway] and not isClosed then
 		if waterway == "river" and way:Holds("name") then
-			way:Layer("water_name_detail", false)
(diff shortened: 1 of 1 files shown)


systemed:
Thank you!

PR #182, opened 2020-09-10
Exclude more basins

leonardehrenfried:
I noticed a few more basins and other artificial water bodies showing up on our map so I modified the selection a little.

If you're interest in what those things are, then you can visit the links in the code.

## Before
Screenshot from 2020-09-10 11-29-45
## After
Screenshot from 2020-09-10 11-30-10
## Before
Screenshot from 2020-09-10 11-23-37
## After
Screenshot from 2020-09-10 11-08-02


systemed:
Looks good - thanks!

PR #197, opened 2021-02-21
Use boost asio to run threadpool

kleunen:
- Small optimization to mmap storage
- Use thread pool to run the tile generation

Can you check if the shapes are rendered correctly, it seems they always output 0 at my machine:
Shape points: 0, lines: 0, polygons: 0


kleunen:
The CI now failes due to boost version. But the CI also has newer boost version. I think you can configure with the BOOST_ROOT define. You actually don't have to install using apt-get.

kleunen:
https://github.com/actions/virtual-environments/issues/319

systemed:
A very handy 20% improvement on the Great Britain extract:

````
real	43m58.630s
user	237m29.682s
sys	14m49.743s
````

Haven't tested the shapefile rendering yet but the output log looks good ("Shape points: 0, lines: 0, polygons: 439").

**Comment by kleunen** [comment]
The CI now has the compiled executable as artifact to download. You can also compile windows and macos executable using actions and have as download. Completely link it static. Can be convenient for users to download

**Comment by kleunen** [comment]
I think it is also possible to make the reading of the pbf somewhat parallel. The decompression can be done parallel

**Comment by kleunen** [comment]
Stored 112737562 nodes, 14982848 ways, 75012 relations
Shape points: 0, lines: 0, polygons: 24
Generated points: 9718722, lines: 4523991, polygons: 12568893
Zoom level 14, writing tile 38000 of 51383
Filled the tileset with good things at netherlands.mbtiles

real    123m31.701s
user    56m30.177s
sys     21m11.627s

For me the real time is twice the time for user (cpu time). 
So the process is waiting many times for the kernel, maybe for reading the file input ?

File: src/tilemaker.cpp [hunk 1]
```diff
@@ -281,24 +290,43 @@ int main(int argc, char* argv[]) {
...
+
+		// Mutex is hold when IO is performed
+		std::mutex io_mutex;
+
 		for (uint zoom=sharedData.config.startZoom; zoom<=sharedData.config.endZoom; zoom++) {
 
-			tileData.SetZoom(zoom);
-			sharedData.zoom = zoom;
+			// Loop through tiles
+			uint tc = 0;
```
systemed, line 302: We should move this outside the loop to be consistent with `total_tiles`, I think?

kleunen: Ah yes, I noticed the counter was wrong, but that is why

kleunen: Zoom level 14, writing tile 38000 of 51383

src/tilemaker.cpp
```diff
@@ -278,14 +284,13 @@ int main(int argc, char* argv[]) {
 			} else if (srcZ > config.startZoom) {
 				cout << "Mapsplit tiles (zoom " << srcZ << ") can't write data at zoom level " << config.startZoom << endl;
 			}
+
 			cout << "Reading tile " << srcZ << ": " << srcX << "," << srcY << " (" << (run+1) << "/" << runs << ")" << endl;
 			vector<char> pbf = mapsplitFile.readTile(srcZ,srcX,tmsY);
-			// Write to a temp file and read back in - can we do this better via a stringstream or similar?
```
kleunen, line 274: I also removed this temporary file, but I did not test this. Can you test this ? It should be OK llike this

systemed:
Thought I'd stress-test it by running it on a France extract (lots of buildings)!

````
real	131m26.129s
user	842m48.936s
sys	28m25.527s
````

14GB memory used, 2.8GB output file, and the map looks great.

**Comment by kleunen** [comment]
But what is this mapsplitting the temporary file was used for ?
I removed the temporary, but I do not know how to test

**Comment by kleunen** [comment]
You can also try complete Europe with zoom level op to 12 or so ?

File: src/tilemaker.cpp (hunk 1 above)
Comment (kleunen) [reply src/tilemaker.cpp]: And I added a print when the final tile is generated.

**Comment by systemed** [comment]
mapsplit is a Java tool to split a single .osm.pbf into a database of tiled .osm.pbfs: https://wiki.openstreetmap.org/wiki/Mapsplit , https://github.com/simonpoole/mapsplit

It's in tilemaker as an option to reduce memory usage - tilemaker reads in a square of source tiles from the .msf file at once, creates the MVT output from them, then loops. Here's a sample .msf file: https://www.systemed.net/temp/centre-latest.msf

It's less crucial than it once was now that we have the mmap store, but I think it's probably worth keeping as it offers a further reduction in memory usage. I've just run this branch over it and it worked fine.

**Comment by kleunen** [comment]
I am trying out 1 more optimization

**Comment by kleunen** [comment]
I tried compiling against luajit, but that actually seems broken.

**Comment by systemed** [comment]
Luajit is famously pernickety (e.g. https://github.com/Project-OSRM/osrm-backend/issues/2955) - I'll give that some attention in due course.

**Comment by kleunen** [comment]
Yeah, well. I guess of normal lua works well, it seems luajit is also not maintained anymore. 

I just wonder if it is possible to run the lua scripts in parallel, such that the generation of geometries becomes parallel. But at the moment I can't say anything about that, i have no clear view of how the generation of the geometries based on lua works. 

I think it is possible to first load all nodes/ways/relations in the OSMStore, and then start executing the lua scripts for each generated node/way/relation ? Or is that too naive to think ?

**Comment by kleunen** [comment]
But to actually process big maps, you still are going to need a lot of RAM. On my 8G machine, i can convert Netherlands now. But you see the process is many times in this D state, waiting for IO. It is not going to 100% cpu, because it is IO bounded now. 

So if you would like to convert the whole planet, you still need a fast SSD and probably around 100G of RAM. That is my guess-timate :)

File: include/osm_store.h [hunk 3]
```diff
@@ -296,7 +309,7 @@ class OSMStore
 				return;
 			} catch(boost::interprocess::bad_alloc &e) {
 				map_size = map_size * 2;
-				std::cout << "Resizing osm store to size: " << map_size << std::endl;
+				std::cout << "Resizing osm store to size: " << (map_size / 1000000) << "M                " << std::endl;
```
systemed, line 312: The actual filesize of the osm_store.dat file appears to be twice the reported value - maybe change 1000000 to 500000?

kleunen: Yes, i noticed this too. I am not sure why this is, I do give this value to boost interprocess grow method. If interprocess allocates 2 * size I specify, i would rather ask for size / 2 there. Because otherwise it would be confusing.

kleunen: https://github.com/systemed/tilemaker/blob/master/include/osm_store.h#L297-L307

kleunen: Oh, wait. It grows with _extra_ bytes, not to the specified size

````C++
   //!Tries to resize mapped file so that we have room for
   //!more objects.
   //!
   //!This function is not synchronized so no other thread or process should
   //!be reading or writing the file
   static bool grow(const char *filename, size_type extra_bytes)
   {
      return base_t::template grow
         <basic_managed_mapped_file>(filename, extra_bytes);
   }

**Comment by systemed** [comment]
> So if you would like to convert the whole planet, you still need a fast SSD and probably around 100G of RAM. That is my guess-timate :)

I'm trying with Europe now (lots of RAM, no SSD) but it's got stuck at block 100663, using 5%ish of CPU. I'll need to do a bit of digging to find why that is - possibly a dodgy polygon or similar.

**Comment by kleunen** [comment]
> > So if you would like to convert the whole planet, you still need a fast SSD and probably around 100G of RAM. That is my guess-timate :)
> 
> I'm trying with Europe now (lots of RAM, no SSD) but it's got stuck at block 100663, using 5%ish of CPU. I'll need to do a bit of digging to find why that is - possibly a dodgy polygon or similar.



> > So if you would like to convert the whole planet, you still need a fast SSD and probably around 100G of RAM. That is my guess-timate :)
> 
> I'm trying with Europe now (lots of RAM, no SSD) but it's got stuck at block 100663, using 5%ish of CPU. I'll need to do a bit of digging to find why that is - possibly a dodgy polygon or similar.

Well, i noticed the same thing. I am converting Netherlands now, with Germany and Belgium to get some nice maps around the border. But now the process is stopped and only doing 2%. 

I see my OSM store is:
16G Feb 22 18:01 osm_store.dat

While the last message from tilemakers was:
Resizing osm store to size: 16384000000

Without the *2 fix, this means the osm_store.dat should be now 32G. But I think linux is actually denying me to allocate the virtual memory space and stalling my process now.

**Comment by kleunen** [comment]
Reading .pbf netherlands-latest.osm.pbf
Resizing osm store to size: 128000000
Resizing osm store to size: 256000000
Resizing osm store to size: 512000000
Resizing osm store to size: 1024000000
Resizing osm store to size: 2048000000
Resizing osm store to size: 4096000000
Resizing osm store to size: 8192000000s 0
Geometry is empty in OsmLuaProcessing::LayerAsCentroid
Can't add Attribute rank if no Layer set
Can't add Attribute class if no Layer set
Can't add Attribute subclass if no Layer set
Geometry is empty in OsmLuaProcessing::LayerAsCentroid
Can't add Attribute rank if no Layer set
Can't add Attribute class if no Layer set
Can't add Attribute subclass if no Layer set
Geometry is empty in OsmLuaProcessing::LayerAsCentroid
Geometry is empty in OsmLuaProcessing::LayerAsCentroid

Stored 112737562 nodes, 14982848 ways, 75012 relations
Shape points: 0, lines: 0, polygons: 24
Generated points: 9718722, lines: 4523991, polygons: 12568893
Reading .pbf belgium-latest.osm.pbf
Geometry is empty in OsmLuaProcessing::LayerAsCentroid
Geometry is empty in OsmLuaProcessing::LayerAsCentroid
Geometry is empty in OsmLuaProcessing::LayerAsCentroid
Geometry is empty in OsmLuaProcessing::LayerAsCentroid
Geometry is empty in OsmLuaProcessing::LayerAsCentroid
Geometry is empty in OsmLuaProcessing::LayerAsCentroid
Geometry is empty in OsmLuaProcessing::LayerAsCentroid

Stored 46716127 nodes, 6706315 ways, 16625 relations
Shape points: 0, lines: 0, polygons: 24
Generated points: 12704191, lines: 6955491, polygons: 17952005
Reading .pbf germany-latest.osm.pbf
Resizing osm store to size: 16384000000

**Comment by kleunen** [comment]
I will try only Germany

**Comment by systemed** [comment]
Germany's fine:

````
real	76m51.204s
user	120m42.150s
sys	23m6.006s
````

**Comment by kleunen** [comment]
I think i can actually improve also the reading speed by making the parsing also parallel. So if rendering europe fails, we can have another go maybe later. Maybe get some more experience on smaller areas. 

I do have the feeling the process sometimes stalls after having to rehash the node store after resizing the mmap. This becomes a very io intensive operation if the hashmap is stored on disk. Cpu might drop to very low levels because of this, it has to wait for the disk before it can continue rehashing.

**Comment by kleunen** [comment]
It was going along quite nicely, my Netherlands/Germany/Belgium conversion. But it got killed by the OOM now:

Feb 23 12:24:29 homeserver kernel: Out of memory: Killed process 1269747 (tilemaker) total-vm:41548992kB, anon-rss:5328644kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:73440kB oom_score_adj:0
Feb 23 12:24:29 homeserver kernel: oom_reaper: reaped process 1269747 (tilemaker), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB

I also noticed that the rehashing of the node/ways map takes a considerable time. So i would like to add an option to pre-initialize the mmap file and hash tables to a certain size, such that this rehashing does not occur.

**Comment by kleunen** [comment]
@systemed Can you do another benchmark on Great-Britain ? And when you run it, initialize the store with the number of nodes/ways you expect. I now do the following on NL/BE/Germany (500M nodes, 75M ways):

tilemaker --init-store 500:75 --input netherlands-latest.osm.pbf --input belgium-latest.osm.pbf --input germany-latest.osm.pbf --output netherlands.mbtiles --process process-openmaptiles.lua --config config-openmaptiles.json

**Comment by systemed** [comment]
With `--init-store 300:50`:

````
real	41m55.698s
user	234m31.710s
sys	14m48.350s
````

So a modest but useful improvement!

**Comment by systemed** [comment]
I've tried running Europe with `--init-store 3000:350` (same machine with 144GB RAM).

The good news is that it no longer hangs - it's currently up to pbf block 383993 (after about 12 hours' running) and still going.

It is running slowly now though: now that it's on ways rather than nodes, it's taking about 3 seconds to process each block. The hard drive is churning busily (iotop shows read 10M/s, write much lower at 400-1000K/s). The mmap file is up to 262144M.

**Comment by kleunen** [comment]
I had a second try with converting Germany/NL/BE on my 8G machine. But it got killed again with out of memory. It is almost complete. 
I have one more idea to improve memory usage, shall I try this and include in this PR ? 

Stored 112737562 nodes, 14982848 ways, 75012 relations
Shape points: 0, lines: 0, polygons: 24
Generated points: 9718722, lines: 4523991, polygons: 12568893
Reading .pbf belgium-latest.osm.pbf
Block 7556 group 0 ways 8000 relations 0
Stored 46716127 nodes, 6706315 ways, 16625 relations
Shape points: 0, lines: 0, polygons: 24
Generated points: 12704191, lines: 6955491, polygons: 17952005
Reading .pbf germany-latest.osm.pbf

Resizing osm store to size: 65536M
Killed50726 group 0 ways 8000 relations 0

real    974m24.789s
user    120m9.914s
sys     177m46.233s

I think germany is about 56k blocks, so I am always there with converting :)

**Comment by systemed** [comment]
> shall I try this and include in this PR ?

Sure!

**Comment by kleunen** [comment]
> I've tried running Europe with `--init-store 3000:350` (same machine with 144GB RAM).
> 
> The good news is that it no longer hangs - it's currently up to pbf block 383993 (after about 12 hours' running) and still going.
> 
> It is running slowly now though: now that it's on ways rather than nodes, it's taking about 3 seconds to process each block. The hard drive is churning busily (iotop shows read 10M/s, write much lower at 400-1000K/s). The mmap file is up to 262144M.

So the hang was related to the rehashing you think ? 
That is quite a massive mmap file :P In the end I had a 65536M on 8G of ram. So you can still stretch it somewhat. But what I see is that if the mmap file gets to big compared to the available RAM. the processing slows down to 30-40% CPU, because it is waiting all the time of mmap blocks to be swapped in/out. :)

You can see that also in my processing time: 974 minutes wall clock time, but only  120 + 177 minutes CPU time. 
So effectively it use using 30% CPU and the rest is just waiting for IO :)

**Comment by systemed** [comment]
Exactly that. CPU is about 10% at the moment (still running).

**Comment by kleunen** [comment]
have you converted Europe before with tilemaker ?

**Comment by systemed** [comment]
No, never.

**Comment by kleunen** [comment]
Unventured territories :D Excellent :D

**Comment by kleunen** [comment]
How is it going with the Europe map generation?

**Comment by systemed** [comment]
Unfortunately it was thrashing the HDD so much that I eventually ctrl-Ced it - I'm on spinning rust rather than SSD so it's pretty slow.

I thought I'd then have a go with a mapsplit file but ran into https://github.com/simonpoole/mapsplit/issues/15 . I'll give it some more thought as to how/whether there might be ways round this, but it's not a massive deal - I wouldn't usually be using an unfiltered .pbf anyway (I don't want buildings and they're 50%+ of the dataset!).

**Comment by kleunen** [comment]
Yes, I think that generating Europe on that machine is simply too much too ask. I made some more small improvements on memory usage, and I still have one more idea I would like to try. If you would like to convert Europe on that machine, you would definitly need a SSD drive, the HDD is simply not going to cut it. 

If you would remove the buildings and renumber the whole of Europe, would the NodeID fit in 32bits (COMPACT_NODES ?). Because that would be a huge memory saving.

**Comment by systemed** [comment]
Renumbering it would fit in 32 bits even with the buildings retained. Maybe I'll try that!

**Comment by kleunen** [comment]
ok. Just wait for couple of days, i may have another update. 
I already committed one optimization, you can try that again on GB ?

**Comment by systemed** [comment]
Will do!

**Comment by systemed** [comment]
Big improvement:

````
real	36m54.930s
user	231m17.416s
sys	2m13.766s
````

Weirdly it segfaulted on exiting the program, though it doesn't appear to have made any difference:

````
Zoom level 14, writing tile 1060157 of 1060157               
Filled the tileset with good things at ../tm_debug/gb_25.mbtiles
Segmentation fault (core dumped)
````

**Comment by kleunen** [comment]
I think this has to do with the intrusive_ptr. It should be locked, but it is not now, so I should add a lock to allow multithreaded delete of the intrusive_ptr

**Comment by systemed** [comment]
Runs fine without segfaulting now.

**Comment by systemed** [comment]
Weirdly the compiler throws this if just building with a standard make:

````
g++ -O3 -Wall -Wno-unknown-pragmas -Wno-sign-compare -std=c++14 -pthread -DCOMPACT_NODES -o src/output_object.o -c src/output_object.cpp -I/usr/local/include -isystem ./include -I./src -I/usr/include/lua5.1
In file included from ./include/geomtypes.h:6,
                 from ./include/output_object.h:9,
                 from src/output_object.cpp:5:
/usr/include/c++/9/limits: In instantiation of ‘static constexpr _Tp std::numeric_limits<_Tp>::max() [with _Tp = std::atomic<unsigned int>]’:
./include/output_object.h:112:2:   required from here
/usr/include/c++/9/limits:321:48: error: use of deleted function ‘std::atomic<unsigned int>::atomic(const std::atomic<unsigned int>&)’
  321 |       max() _GLIBCXX_USE_NOEXCEPT { return _Tp(); }
      |                                                ^
In file included from /usr/include/boost/smart_ptr/detail/sp_counted_base_std_atomic.hpp:21,
                 from /usr/include/boost/smart_ptr/detail/sp_counted_base.hpp:48,
                 from /usr/include/boost/smart_ptr/detail/shared_count.hpp:29,
                 from /usr/include/boost/smart_ptr/shared_ptr.hpp:28,
                 from /usr/include/boost/shared_ptr.hpp:17,
                 from /usr/include/boost/date_time/microsec_time_clock.hpp:18,
                 from /usr/include/boost/interprocess/detail/posix_time_types_wrk.hpp:31,
                 from /usr/include/boost/interprocess/detail/os_thread_functions.hpp:36,
                 from /usr/include/boost/interprocess/detail/managed_open_or_create_impl.hpp:23,
                 from /usr/include/boost/interprocess/managed_mapped_file.hpp:24,
                 from ./include/geomtypes.h:20,
                 from ./include/output_object.h:9,
                 from src/output_object.cpp:5:
/usr/include/c++/9/atomic:778:7: note: declared here
  778 |       atomic(const atomic&) = delete;
      |       ^~~~~~
make: *** [Makefile:76: src/output_object.o] Error 1
````

but with cmake it's fine.

**Comment by kleunen** [comment]
I enabled c++14 in the cmake file.
Move back to c++11 or enable c++14 in the makefile also?

**Comment by systemed** [comment]
Oddly it does the same with c++14 but it's fine with c++17!

**Comment by kleunen** [comment]
Ah, now I remember. It was this max check, i removed this already.

**Comment by leonardehrenfried** [comment]
What currently fails in CI is the Docker image build as it uses the Makefile and not CMake. Should it use CMake?

**Comment by leonardehrenfried** [comment]
Also, in the previous PR @kleunen said that it would be useful to use a larger geographic extract in CI to compare run times. Is that still the case?

**Comment by kleunen** [comment]
> What currently fails in CI is the Docker image build as it uses the Makefile and not CMake. Should it use CMake?

If you just want to test the branch yes. I think I will push another commit in hour or so. If you want to test, you can wait. The additional commit contains some additional memory savings.

**Comment by leonardehrenfried** [comment]
I can definitely wait but I was wondering if CMake is the correct way to build or if both are supported.

**Comment by kleunen** [comment]
> I can definitely wait but I was wondering if CMake is the correct way to build or if both are supported.

Both are supported, i just broke the makefile in this PR :)

**Comment by kleunen** [comment]
It is building again. 

@systemed Can you test again on the Great-Britain conversion ? I think it may actually be a little bit slower now. But the memory saving is quite big. I redid the Netherlands conversion and it is a lot faster on my memory constrained machine. I think with 32bit node ids you may actually be able to convert Europe with this on your machine. 

Have a test if the attributes also work correctly. Also with mapsplit.

**Comment by systemed** [comment]
Great Britain:

````
real	41m46.774s
user	230m18.140s
sys	2m0.130s
````

Peak RAM used was around 5.2GB. I'll do some more testing...

**Comment by systemed** [comment]
No different with Europe I'm afraid (using a renumbered version and COMPACT_NODES) - it basically starts churning the hard disk at around block 380000, midway through reading the ways. RAM usage is still low at around 10.4GB. 

I wonder if there might be some mileage in having a special, optional mode for renumbered nodes, where we could just use a plain vector NodeStore rather than an unordered_map (because node IDs are consecutive). But that's probably one for another PR!

**Comment by kleunen** [comment]
> I wonder if there might be some mileage in having a special, optional mode for renumbered nodes, where we could just use a plain vector NodeStore rather than an unordered_map

Yes i thought about this too. But i dont think this will make a difference. You can adjust the load factor which changes when the hash map is resized. But this value is actually quite high. Dont expect the hash map to be very empty. It is probably just about as full as a vector.

What i do notice is that linux may swap out the mmap blocks very early. And has to reload it again. Why is only 10g used if much more memory is available. It should keep the mmap in memory for as long as possible.

**Comment by systemed** [comment]
htop is showing RES 92.2G so I guess it's mapping a fair amount into memory. I wonder whether the WayStore has grown to such a size that it can't keep the full NodeStore in memory any more.

**Comment by kleunen** [comment]
> htop is showing RES 92.2G so I guess it's mapping a fair amount into memory. I wonder whether the WayStore has grown to such a size that it can't keep the full NodeStore in memory any more.

Yes. The node store is too big. And when you are processing a way, the nodes need to be looked up in the nodes map. And this causes pages to be swapped in and out.

Ideally you want related nodes to be stored in the same page on the drive. But i think it does. The hash function actually keeps the ids sequential and does not completely randomize them. So they should be in buckets close to each other.

The generated ways get swapped out quickly to disk. Only the nodes are read very often when loading the pbf, because of the ways. And if this does not fit in memory. It needs to be loaded from disk everytime. And the way building stalls

**Comment by kleunen** [comment]
An hdd is simply not suited for this behaviour. The process performs random reads and the hdd needs to seek the block. Ssd has much better random read performence.

**Comment by kleunen** [comment]
Ah wait. The vector does not need to store the key and the hashmap does. So this indeed will save memory

**Comment by kleunen** [comment]
> Ah wait. The vector does not need to store the key and the hashmap does. So this indeed will save memory

Does osmium renumber say how much nodes and ways there are in Europe?

**Comment by systemed** [comment]
2710803032 nodes, 325100853 ways. (From `osmium fileinfo`. osmium is awesome.)

**Comment by kleunen** [comment]
````
struct LatpLon {
    int32_t latp;
    int32_t lon;
};
````

So this is 8 bytes per Node, and NodeID = 4 bytes. But I think actually, unordered_map aligns the entries on 8 bytes anyway. Because unaligned read is slow. So if you set the hashmap to 3000 nodes preallocated, you would need 3000 * 16 bytes = 48G for the node store. So this should fit in the memory. 

If you preallocate a vector, with 32bits node ID, 3000 entries, 3000 * 4 bytes = 12G. It is worth a try, 12G should be easy to keep in memory.

**Comment by kleunen** [comment]
I also though about you can use ZSTD compression on the ways with a dictionary:
https://facebook.github.io/zstd/

So if you have a linestring or polygon, with a sequence of node ids, you first substract every node id from previous node id:
id1 id2 id3 id4
becomes:
id1 (id2 - id1) (id3 - id2) (id4 - id3)

This will usually result in a sequence like:
id1 1 1 1

But I think with a fast enough SSD, it will already be easy to keep 1 CPU busy and you may be able to perform a parallel loading actually. Use multiple CPU when loading the PBF.

**Comment by kleunen** [comment]
What I see on IOtop now. My tilemaker is using 32G on NL+BE+DE conversion. And first it reads with 120Mb/sec from this. Does some processing. And then you see a write burst with 80Mb/sec.

**Comment by kleunen** [comment]
Maybe make a btrfs partition with zstd compression enabled, and put the mmap file on this. This may improve the read/write performance:
https://linuxhint.com/enable-btrfs-filesystem-compression/#:~:text=ZSTD%3A%20ZSTD%20or%20Zstandard%20is,the%20ZSTD%20data%20compression%20algorithm.

**Comment by kleunen** [comment]
These algorithms are actually causing the ways to be loaded into memory again:
https://github.com/systemed/tilemaker/blob/master/include/osm_store.h#L422-L526

So if memory is limited, the ways are loaded, nodes being pushed out of memory. Later nodes being loaded again into memory, ways pushed out. Etc ... :)

**Comment by kleunen** [comment]
I converted now Netherlands with the edges around the borders with germany and belgium. I first did a bounding box export from Europe. And then converted using tilemaker:

Stored 176982865 nodes, 25353999 ways, 94603 relations
Shape points: 0, lines: 0, polygons: 24
Generated points: 14316834, lines: 8192207, polygons: 20968705
Zoom level 14, writing tile 66000 of 66000
Filled the tileset with good things at netherlands.mbtiles

real    193m54.834s
user    83m2.147s
sys     30m3.398s

I have 8G of main memory and 4G swap. This works quite well, it is possible now. It is IO bounded when performing the operation, but still, the conversion time is quite acceptable. What I do notice is that the final file is not completely clipped to the bounding box. I have some parts which are extending out of the bounding box. 

https://mqtt.kleunen.nl/map/

**Comment by kleunen** [comment]
Here I am missing a street ? 
image

vs OSM:
image

**Comment by systemed** [comment]
I've experimented with using a sequential node store.

In osm_store.h the quick-and-dirty changes are:

````C++
template <typename T> using scoped_alloc_t = boost::container::scoped_allocator_adaptor<T>;
using ll_allocator_t = boost::interprocess::node_allocator<LatpLon, mmap_file_t::segment_manager>;
using ll_vector_t = boost::container::vector<LatpLon, scoped_alloc_t<ll_allocator_t>>;
using vector_t = boost::container::vector<LatpLon, scoped_alloc_t<ll_allocator_t>>;
vector_t *mLatpLons;

void reopen(mmap_file_t &mmap_file)
{
    mLatpLons = mmap_file.find_or_construct<vector_t>("node_store")(mmap_file.get_segment_manager());
}

void reserve(uint nodes) {
    mLatpLons->reserve(nodes);
    mLatpLons->resize(nodes);
}
````

and then in osm_store.cpp:

````C++
size_t NodeStore::count(NodeID i) const {
    return mLatpLons->size() > i ? 0 : 1;
}

void NodeStore::insert_back(NodeID i, LatpLon coord) {
    mLatpLons->at(i) = coord;
}

void NodeStore::clear() {
// commented out because otherwise it resizes back to 0
//	mLatpLons->clear();
}
````

And... it appears to work! Slowly. I set it running last night and it has literally just finished reading in the .pbf. There was a lot of HDD churn in the latter stages, but it got much further into the process before that started. It's now just starting to write tiles:

````
Zoom level 5, writing tile 100 of 22946608
````

Memo to self: I need to specify 4200 as the size for the nodestore, not 4300 as that overflows the uint. ;)

How best to integrate this as an option I'm not sure - I'm guessing some template magick would be the right approach but that starts to stretch my C++ knowledge...

----

Netherlands map looks great! I think the missing road is a result of the highway=construction tag - we do handle this in the Lua profile but possibly not in the way that the OMT stylesheet expects.

ZSTD-like delta encoding for nodelists would be an interesting idea and could save a lot of memory. That's exactly how the .pbf format works (https://wiki.openstreetmap.org/wiki/PBF_Format#Ways_and_Relations). I briefly looked for an existing C++/C implementation we could drop in but haven't turned anything up yet.

**Comment by kleunen** [comment]
Yes, well, actually, i was thinking about Fibonacci Coding to store Linestrings and Rings:
https://www.geeksforgeeks.org/fibonacci-coding/

I think we can win a lot there. And the Fibonacci coding is easier than importing a complete compression library. Although with compression would also be possible. Because actually, you are generating a stream of linestring/rings. You never remove a linestring from the mmap file. So basicly you are just appending them to the back of the file. But still, coding only on a vector might be easier. 

Also, if you use COMPACT_NODES, you can use a vector on the node id, like you did. And also compress the linestring/polygons with fibonacci coding. Because the nodeid's are smaller, the difference between two nodeids will also be smaller. So they will be compressed efficiently using fibonacci coding.

**Comment by kleunen** [comment]
You won't need any template magic. You can just implement two node stores: one using the unordered_map and the other using the vector. You can switch with the macro COMPACT_NODES between the implementation. The OSMStore actually only uses the reopen/reserve/at and insert_back. You can make a base class for this with virtual methods, and have two implementations. Using a commandline option you can then switch between normal/compact mode. And create the Normal or Compact node store from main.cpp and give a pointer to OSMStore. That way you can actually configure it using a command line option.

**Comment by kleunen** [comment]
How big is your swapfile on the system btw?

**Comment by kleunen** [comment]
> ZSTD-like delta encoding for nodelists would be an interesting idea and could save a lot of memory. That's exactly how the .pbf format works

Yes. If you keep it like it was in the original file. You dont even need to recode the nodelist. Maybe just need to unpack it using an iterator when you use it.

**Comment by systemed** [comment]
osm_store.dat is currently 256GB. It's currently on z9 and the HDD is spinning so much it's about to go into orbit. I've just ordered an SSD.

Do you think this PR is ready to merge as is or would you like to do some more to it?

**Comment by kleunen** [comment]
No. You can merge like this. :)

Is it using 24 threads now? Or not? Was the ssd expensive? Because you need a sas ssd right? :)

**Comment by systemed** [comment]
SAS would have been nice but way too expensive! I've gone for SATA for now - I figure it'll still be a big improvement over currently. I'll rerun Europe when it arrives: for now I've killed it because most of the threads were pretty much waiting for IO.

**Comment by kleunen** [comment]
Yes. Sata is much better deal. I thought the hp only had sas. It will make a difference yes.

**Comment by systemed** [comment]
It'll do SATA but at 3Gbps rather than 6Gbps. Still much better than what I'm getting currently!

Now merged - thanks for another awesome PR.

**Comment by leonardehrenfried** [comment]
I tried this on a box with 64GB of memory and an SSD. The whole processing of Germany was actually a little slower than before, presumably because it is not keeping everything in memory anymore.

Is it possible to tell tilemaker to keep everything in memory, as it did before? Some comments mention a RAM disk or a tmpfs but I cannot find any information about specifying where the mmap file should be stored.

**Comment by kleunen** [comment]
````
  --store arg (=osm_store.dat) temporary storage for node/ways/relations data
  --init-store arg (=20:5)     initial number of millions of entries for the
                               nodes (20M) and ways (5M)
  --verbose                    verbose error output
````

You can also initialize the store with the number of nodes/ways that you expect. Did you really time the conversion ? 
Because we have not seen a slow-down.

**Comment by leonardehrenfried** [comment]
I'm comparing against the code before you introduced the mmap (not to before the asio) and I noticed perhaps a 5% slowdown.

Thanks for the command line params. I will try them and report back.

**Comment by leonardehrenfried** [comment]
So I tried putting the mmap file into `/dev/shm` which in Debian systems is a `tmpfs`. The full command is this:
```
tilemaker
  --output=/var/tilemaker/output.mbtiles \
  --config=/var/tilemaker/config-openmaptiles.json \
  --process=/var/tilemaker/process-openmaptiles.lua \
  --store=/dev/shm/tilemaker-osm.dat \
  --verbose
```
However, the process exits with code 135 after opening the pbf file without having processed anything.

The log is the following:

```
Layer place (z4-14)
Layer poi (z12-14)
Layer poi_detail (z14-14) -> poi
Layer housenumber (z14-14)
Layer waterway (z6-14)
Layer waterway_detail (z12-14) -> waterway
Layer transportation (z8-14)
Layer transportation_main (z9-14) -> transportation
Layer transportation_mid (z11-14) -> transportation
Layer transportation_detail (z13-14) -> transportation
Layer transportation_name (z8-14)
Layer transportation_name_mid (z12-14) -> transportation_name
Layer transportation_name_detail (z14-14) -> transportation_name
Layer building (z13-14)
Layer water (z6-14)
Layer ocean (z6-14) -> water
Layer water_name (z13-14)
Layer water_name_detail (z13-14) -> water_name
Layer aeroway (z11-14)
Layer aerodrome_label (z10-14)
Layer park (z11-14)
Layer landuse (z11-14)
Layer landcover (z6-14)
Layer mountain_peak (z11-14)
Initializing storage to 20M nodes and 5M ways
Reading .shp ocean
Reading .pbf /srv/tilemaker-input.osm.pbf
Unable to open coastline/water_polygons.shp or coastline/water_polygons.SHP.
```

Do you know why this happens?

leonardehrenfried:
Hmm, if it configure the mmap file to placed in `/tmp` which on Debian is not a `tmpfs` it seems to work. Do you have an idea why it could be problematic with `tmpfs`?

leonardehrenfried:
Ah, hang on. I'm running this in a docker container so perhaps in there `/dev/shm` doesn't work...

kleunen:
I have a new laptop with intel core i7 10850 and 32G of ram. I am converting now 1/4 of the planet (south america). But what I see now is that when I get to the higher zoom level, the cpu usage drops to 33% and tilemaker is just generating the tiles very fast. I think we have to start a work item on the thread pool for every 100 tiles instead of every single tile. Because I think the conversion is just going to fast to keep all my cpu's busy (12) and lock contention occurs when starting the work items on the thread pool.

PR #198, opened 2021-02-28
VarInt encoded waylist and relations

kleunen:
I actually implemented the VarInt ended ways/relations. I just wanted to give it a try :D It seems it is slower when having plenty or RAM. I will try now on my Netherlands extract :)

### DIFF
--- a/include/osm_lua_processing.h
+++ b/include/osm_lua_processing.h
@@ -180,8 +180,8 @@ private:
 	bool isWay, isRelation;					///< Way, node, relation?
 
 	int32_t lon1,latp1,lon2,latp2;			///< Start/end co-ordinates of OSM object
-	NodeVec *nodeVec;						///< node vector
-	WayVec *outerWayVec, *innerWayVec;		///< way vectors
+	OSMStore::handle_t nodeVec;				///< node vector
+  	WayVec *outerWayVec, *innerWayVec;      ///< way vectors
 
 	Linestring linestringCache;
 	bool linestringInited;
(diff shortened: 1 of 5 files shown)


kleunen:
With VarInt on my Netherlands Conversion:
real    207m12.466s
user    86m17.450s
sys     32m0.442s

Without:
real    193m54.834s
user    83m2.147s
sys     30m3.398s

So no improvement yet :)

kleunen:
After optimization:
real    173m44.571s
user    82m10.237s
sys     25m23.959s

kleunen:
Using COMPACT_NODES has a much bigger impact:
real    63m41.214s
user    66m43.087s
sys     6m56.317s

systemed:
I'll run it for the UK and see how what time it gives...

kleunen:
Ok, you are free to try. But to be honest, I find the approach too much extra complexity for the benefit it gives. I would recommend to add a "compact mode. Where the normal mode is:

64 bits Node ID
Node store: unordered_map
Way store: 64 bit Node ID's

and Compact mode (32 bit Node ID):
Node store: preallocated vector
Way store: 32 bit id's. 

Would be much more straightforward to implement, and I notice a very big speedup with osmium renumber + 32 bits NodeIDs on memory constrained system.

systemed:
Great Britain (no init-store argument, 64-bit node IDs):

````
real	38m11.607s
user	229m55.563s
sys	2m8.412s
````

So there's a modest improvement again but nothing massive. I'll keep this open for now but yes, COMPACT_NODES is probably the right thing to focus on.

kleunen:
> So there's a modest improvement again but nothing massive. I'll keep this open for now but yes, COMPACT_NODES is probably the right thing to focus on.

I will have a look if I can make a command line configurable compact mode. I don't have as much spare time as last week, so, will take a bit longer to finish :)

If you have your SSD i am very much interested in the results of the Europe conversion :)

systemed:
I'll let you know when it arrives!

PR #199, opened 2021-03-04
Allow tilemaker to run in compact (32bit) nodeid mode

kleunen:
This PR gives the ability to switch to allow tilemaker to switch between 32bit and 64bit node id mode from the command line. This adds the following command line option:

````
  --compact                    Use 32bits NodeIDs and reduce overall memory
                               usage (compact mode).
                               This requires the input to be renumbered and the
                               init-store to be configured
````

It gives the following notice when running in this mode:
````

Important: Tilemaker running with 32-bit node support.
Use 'osmium renumber' first if working with OpenStreetMap-sourced data,
initialize the init store to the highest NodeID that is stored in the input file.

````

It hugely improves the conversion time of my Netherlands extract on my 8G machine. From something like 200 minutes to 50 minutes. 

I am still thinking about adding the VarInt ways to this patch to further reduce memory consumption. For large conversion (like Europe / Planet), I think every byte counts. So it still can be useful.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,14 +11,14 @@ env:
 jobs:
   build:
     name: Compile, install and build mbtiles
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
     - uses: actions/checkout@v2
     - name: Install dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y --no-install-recommends build-essential liblua5.1-0 liblua5.1-0-dev libprotobuf-dev libsqlite3-dev protobuf-compiler shapelib libshp-dev 
+        sudo apt-get install -y --no-install-recommends build-essential liblua5.1-0 liblua5.1-0-dev libprotobuf-dev libsqlite3-dev protobuf-compiler shapelib libshp-dev libboost-all-dev
 
     - name: Build and install
       env:
@@ -29,7 +29,7 @@ jobs:
         mkdir build
         cd build  
         export CFLAGS=${S_CFLAGS} && export CXXFLAGS=${S_CXXFLAGS} && export LDFLAGS=${S_LDFLAGS}
-        BOOST_ROOT=$BOOST_ROOT_1_72_0 cmake -DTILEMAKER_USE_STATIC_BOOST=true ..
+        cmake -DTILEMAKER_USE_STATIC_BOOST=true ..
         make -j 
         sudo make install
 
@@ -46,7 +46,7 @@ jobs:
 
   Github-Action:
     name: Generate mbtiles with Github Action
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
     - name: Check out repository
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.0)
 
 if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE "Release")
--- a/include/geomtypes.h
+++ b/include/geomtypes.h
@@ -132,16 +132,9 @@ BOOST_GEOMETRY_REGISTER_LINESTRING(mmap::linestring_t)
 BOOST_GEOMETRY_REGISTER_RING(mmap::ring_t)
 BOOST_GEOMETRY_REGISTER_MULTI_POLYGON(mmap::multi_polygon_t)
 
-#ifdef COMPACT_NODES
-typedef uint32_t NodeID;
-#else
 typedef uint64_t NodeID;
-#endif
-#ifdef FAT_WAYS
 typedef uint64_t WayID;
-#else
-typedef uint32_t WayID;
-#endif
+
 #define MAX_WAY_ID std::numeric_limits<WayID>::max()
 typedef std::vector<NodeID> NodeVec;
 typedef std::vector<WayID> WayVec;
--- a/include/osm_lua_processing.h
+++ b/include/osm_lua_processing.h
@@ -180,8 +180,8 @@ private:
 	bool isWay, isRelation;					///< Way, node, relation?
 
 	int32_t lon1,latp1,lon2,latp2;			///< Start/end co-ordinates of OSM object
-	NodeVec *nodeVec;						///< node vector
-	WayVec *outerWayVec, *innerWayVec;		///< way vectors
+	OSMStore::handle_t nodeVec;				///< node vector
+  	WayVec *outerWayVec, *innerWayVec;      ///< way vectors
 
 	Linestring linestringCache;
 	bool linestringInited;
--- a/include/osm_store.h
+++ b/include/osm_store.h
@@ -15,6 +15,8 @@ namespace geom = boost::geometry;
 #include <boost/container/vector.hpp>
 #include <boost/container/scoped_allocator.hpp>
 #include <boost/unordered_map.hpp>
+#include <boost/filesystem.hpp>
+
 //
 // Views of data structures.
 //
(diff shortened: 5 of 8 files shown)


kleunen:
They ditched boost from the CI virtual environments -_-

systemed:
Great Britain with `--compact --init-store 150:25`:

````
real	31m11.341s
user	225m12.862s
sys	1m25.690s
````

🚀

kleunen:
Impressive

carlos-mg89:
I'm willing to do some testing with this PR for the Europe PBF extract.

It happens to be that I finished the processing for the whole Europe, in around 36h. If I could reduce it to 15-20h, it would already be a huge improvement! Although I think that this PR will have bigger impact in smaller PBFs, and not in the bigger ones. I've just taken a quick look to the code, and it looks like this speeds up the tile generation, and not the PBF reading, which from what I was able to understand in the log, is what took me the most (around 30-31 hours), compared to the tile generation (4-5 hours).

If you could point me out what configuration should I pass for the NodeIDs setup, and if possible, what would be the exact command with `osmium renumber` to adapt the `europe.osm.pbf` so it's compatible with these changes.

I intend to run the MBTiles generation once a month, so it'll definitely help me greatly, as the environment.

By the way, @systemed , I give you my most sincere thank you. Before I used the openmaptiles generator, but this greatly simplifies the process, and it actually has sped it up greatly.

kleunen:
The command for renumbering would be: 

````
osmium renumber europe.osm.pbf --output europe-renum.osm.pbf
````

For conversion you would use:
````
tilemaker --compact --init-store 400:100 ...
````

The input is still not parallel no. If you have enough memory in the machine to do the conversion anyway, I am not sure if you would see the improvements you are hoping for. It is mostly focussed on reducing memory now, not improving CPU.

carlos-mg89:
How much memory do you think it could save me up?

My current setup is 32GB DDR4 with a 40GB swap file on a 1TB SSD (7.000MBs
/ 5.000 MBs read/write speed).

The process has taken increasingly the RAM until getting to 32 GB, plus
20-25 GB of the swap file.

I think it'll speed up the process, cause if it makes using less RAM, it'll
use less swap space, and the RAM is much faster.

Missatge de Wouter van Kleunen <notifications@github.com> del dia dv., 5 de
març 2021 a les 13:04:

kleunen:
I guess in that scenario it would give a speedup. In the compact mode, the node positions are possible to stay in memory. I think we calculated it to use about 24G for the europe conversion in compact mode. So this can give a significant speedup. If you can have a look at the conversion process, whether the process stays around 100% CPU and see what can of IOtop throughput you are seeing (read/write mb/sec). That would be helpful.

systemed:
I'm running Europe at the moment and htop is reporting 50.5GB memory used. (VIRT is 296GB, osm_store.dat is 256GB on SSD.) That's using a renumbered extract, and running

````
time tilemaker --input /media/data1/planet/europe-renumbered.osm.pbf --output /media/data3/europe.mbtiles --compact --init-store 3000:100 --store /media/ssd/osm_store.dat --process resources/process-openmaptiles.lua --config resources/config-openmaptiles.json
````

Hopefully should be finished some time this afternoon - I'll report back then.

systemed:
Finished!

````
Block 421500 group 0 ways 0 relations 3625        
Stored 2710803033 nodes, 307809506 ways, 2661947 relations
Shape points: 0, lines: 0, polygons: 4645
Generated points: 91453818, lines: 151714940, polygons: 221195316
Zoom level 14, writing tile 22922705 of 22922705               
Filled the tileset with good things at /media/data3/europe.mbtiles

real	1212m24.417s
user	4212m15.420s
sys	83m53.050s
````

HP DL380 G7, 144GB RAM, 1TB SSD for osm_store.dat, 16 threads. It would drop out now and then to wait for I/O, but was running well across all threads for a majority of the tile-writing stage.

I suspect that a small number of complex geometries are having a disproportionate impact on the processing time, and this might be something I could look at in the future.

kleunen:
Yes, and having some parallel pbf reading process. Or is you CPU now maxed out ? What kind of CPU load an IO speed do you see in the PBF loading and geometry generation step ? Do you think there is still room to keep more than 1 CPU busy ?

systemed:
CPU sits at 100% throughout PBF reading. So I could definitely see parallelisation helping here.

kleunen:
> Decreasing the RAM is wonderful anyway. So I could do the same on the planet, possibly after adding 32GB more of RAM.

That should be possible, with an additional 32G RAM. But having more RAM would be prefered I think, around 128G is maybe more reasonable. But I think you can manage, because you have fast SSD.

kleunen:
> CPU sits at 100% throughout PBF reading. So I could definitely see parallelisation helping here.

Any idea on the IO speed you are seeing when reading the PBF ?

systemed:
It isn't even troubling iotop, which is mostly showing 0.00 B/s read (but some writing, out to osm_store.dat presumably). htop's status column is mostly R, very occasionally S, no D.

kleunen:
I see. Then the next step would be to parallelize the PBF reading. Hmm, not sure how difficult that would be.

systemed:
I _think_ it should be possible to read in a PBF block and then hand it off to a thread pool to be parsed. Probably the main gotcha is ensuring that all nodes have been read before embarking upon ways (and similarly for relations) because of dependencies. But PBF files are ordered so that shouldn't be an issue.

kleunen:
> Probably the main gotcha is ensuring that all nodes have been read before embarking upon ways (and similarly for relations) because of dependencies.

Yes, that is a difficulty. And also, the writing of the output data; OSM data in OSM store, generated geometries, output objects and tiles, this all should be done thread safe. I was thinking maybe it is possible to make some kind of "pipeline", where the PBF blocks are parsed and the generated data is cached, and then finally only a single thread performs the actual storage in OSMStore/Tiles/Output objects. 

But some significant re-architecturing might be required.

systemed:
One to look at later maybe but probably not a priority for now!

@carlos-mg89 That's great that you're finding tilemaker useful - my main rationale for the project was to make it easy for indie devs to create vector tile maps without having to pay Mapbox $$$ or set up a complex database-backed toolchain.

kleunen:
Yes, that was also my motivation :D. But I would need a full planet render, to be able to host the maps myself. I guess I was a bit naive thinking this would be possible on my 8G machine :D I thought, if I give it enough processing time, I would be able to pull it off, but no .. This is outside the capabilities of this poor machine. I have grossly underestimated the planets scale :D And I think the maps outside Europe are not as detailed as in Europe, so possibly there is more too come in the future ? 

But it would be nice if I could start a reasonable AWS instance and render the planet within 24 hours. I think that would not be too much to ask, is it ? :D

systemed:
That would be good!

I'd hoped that mapsplit would be the solution for cases like that - you'd need enough RAM to split the planet, but then tilemaker would have fairly modest demands after that. But it's currently limited to 2^32 nodes which rules out running it on the planet. You could probably pull off a similar trick by using `osmium extract` to split on tile boundary lines (maybe at z4) and running tilemaker for each one with `--merge` and a bounding box, though again, you'd still need enough RAM to run osmium.

kleunen:
Or maybe I can have a talk with SNT of University of Twente, i studied there, i did my PhD there and I still have many contacts there. I know they have a osm.pbf mirror:

https://ftp.snt.utwente.nl/pub/misc/openstreetmap/

But maybe they are inclined to also provide a weekly render of the planet on there infrastructure. They are always inclined to help open projects. There is also the ITC Faculty of Geo-Information Science and Earth Observation. Would be nice if this was free to download for everyone.

kleunen:
> But it's currently limited to 2^32 nodes which rules out running it on the planet

No, you are not limited to 2^32 nodes with this. I was a bit mistaking, but the index in the node store, it can be 64 bits. And when you use the varint encoded ways, they are decompressed to 64 bits. So there is actually nothing holding you back to rendering the planet in this compact mode. 

I also changed the message tilemaker gives at startup, there is no mention of 32bits anymore.

carlos-mg89:
@kleunen have you tried to process the world in batches (with tilemaker) and later to merge all the generated .mbtiles?

This way you won't need to get an AWS instance. This is what I did before generating the MBTiles with tilemaker. I ran the openmaptiles creator for each country of Europe, and then I merged all the MBTiles in a bigger one. And it really works well.

To merge them, I use the tippecanoe tool: `tile-join -o europe.mbtiles *.mbtiles`

The time to generate them would most likely be longer, but you will be able to do it in a short RAM computer.

systemed:
Yes, sorry, the ~2^32~ 2^31 limit is with mapsplit: https://github.com/simonpoole/mapsplit/issues/15 . It's a Java limitation (maximum array size).

systemed:
> To merge them, I use the tippecanoe tool: `tile-join -o europe.mbtiles *.mbtiles`

Oh, that's a good suggestion. I'm planning to do a tilemaker doc on running on resource-constrained systems, so I'll include that.

carlos-mg89:
It works really good, and the merge of MBTiles is quite fast. Yesterday I did it in about 0.5-1 hour (didn't really count). For all separated Europe MBTiles. 20+ files.

kleunen:
Yes, that is a good suggestion. But I was afraid I might get some artifacts on the map from merging, or is this not going to happen ? 

Other thing is, i have a PhD in Computer Science, i have been programming C++ for 20 years now, and I am actually a lecturer on software engineering and C++ on university of applied science. Clearly I am not going to let my self chicken out of a challenge of letting a computer handle multiple GB of geo data :D

No, all bravour aside, i thought it was a not challenge to see if I could scale it up. I liked the approach of the project and I hope it becomes useful too many more people now, not having these high specs machines at their disposal.

kleunen:
> It works really good, and the merge of MBTiles is quite fast. Yesterday I did it in about 0.5-1 hour (didn't really count). For all separated Europe MBTiles. 20+ files.

Yes, I guess on your machine you can render the full planet now, with the approach you specified. Render the different continents and merge them together.

carlos-mg89:
I agree with you that if the process would be optimized it's much better for everyone. Less resources, less money, less consumption of energy, less CO2 and less time.

I could definitely try to help with what I can. Right now, mostly testing. But I could for sure be of some more help. Do you have any kind of roadmap for the project or are you willingly accepting any kind of PR that does improve the project in any way?

kleunen:
You have to ask @systemed, he is the maintainer of the project :) 

I can give some suggestions, I think if you provide feedback and perform testing, this is also useful. Suggestions you give how to use the tools effectively, like using tile-join, it might be more valuable than you think. It can help other people from having to invest a lot of time figuring out how to combine these tools. 

I think the project could also use more styles for rendering. I really have no designer skills, but that could be useful as well :)

kleunen:
What tool does geofabrik use to split out the countries ? 
Of mapsplit does not work on that scale.

kleunen:
Maybe I will give a try in rendering the Benelux (Belgium, netherlands, luxemburg) with the approach @carlos-mg89. Maybe include france, see if I can pull that off on my machine now.

systemed:
Roadmap-wise, I'm aiming for a 2.0 release that enables the average user to spin up vector tiles as easily as:

````
tilemaker great-britain-latest.osm.pbf great-britain.mbtiles
tileserver great-britain.mbtiles
````

To get there, I think the steps are essentially:

- [x] reduce memory usage
- [x] improve performance
- [x] get OpenMapTiles schema up to scratch
- [ ] fix macOS build
- [ ] make installation simpler:
  - [ ] install OMT profile/config into`/usr/local/etc` so you don't have to specify them on the command line
  - [ ] make server.rb into an executable (and install its dependencies)
- [ ] make running simpler:
  - [ ] deduce --input and --output from file extensions if not specified
  - [ ] guess approximate --init-store size from .pbf file size if not specified
  - [ ] tidy logging
- [ ] additional styles, particularly osm-cartographic
- [ ] improve documentation, particularly options for running on constrained systems

@kleunen's recent PRs are a massive step towards this!

But I'm not too prescriptive - if someone has a great idea for something they'd like tilemaker to do, I'm all ears.

systemed:
> What tool does geofabrik use to split out the countries ?

I believe Geofabrik use `osmium extract`. But they have some heavy-duty hardware to call on. Once we've got everything merged and a few tweaks to the OMT schema, I'll ask Frederik to run tilemaker on the full planet. 😄

kleunen:
Well, regarding the roadmap. I don't want to be rude or anything, but I found the ruby tileserver not to be very fitting in the project. I already had lua installed, but for running the conversion you need to install lua, which is a small install. But then for the tileserver, you need to install ruby. I tried this, but what ever i tried, i could not get the ruby tileserver working. I ended up using tileserver-php, which was much easier to setup. 

I think in the end a php based or completely file based tile server would be the best way to go. I think something hosters might have PHP, but not allow SQLite. At least my hoster was replying I needed to convert the sqlite to mysql first, because somehow they only allow certain extentions to be opened locally by PHP, and .mbtiles was blocked. *sigh*

I think a small lua tileserver may be useful for testing out if the conversion worked. But I think the strength of this project is really the conversion, and maybe the tileserver should be left to other projects.

systemed:
Yes, the tileserver really is just there for testing - it's certainly not a production-ready server. It's just so that you can check your tiles look ok before deploying them in your app/site/whatever. Doing it in Lua (probably using pegasus) wouldn't be a daft idea.

kleunen:
And regarding the initial value for the init-store. For the compact it is now required, but it is actually possible to allow the vector to scale beyond the initial configured size. That way it is only a suggestions to improve performance.

systemed:
Do you think this is ready to merge now? Looks good to me.

kleunen:
> Do you think this is ready to merge now? Looks good to me.

Yes. I think it is ok like this

systemed:
Superb. Thank you for another great PR.

tomvanenckevort:
Really interesting to see all the recent performance improvements. Keep up the good work!

One question I do have (and apologies if this isn't the right place) is how much CPU (and the number of CPU cores) has an influence on performance? 
As it's quite easy these days to spin up a powerful VM on Azure or AWS to let Tilemaker do it's thing, would it be beneficial to choose one with many cores, or is there currently a limit on what it can handle?

kleunen:
#204 Gives a good overview of what is possible now given the current state of tilemaker.

Issue #200, opened 2021-03-06
Some POIs are missing.


toshelp:
It worked as expected with the following modifications to process-openmaptiles.lua.

```
diff --git a/resources/process-openmaptiles.lua b/resources/process-openmaptiles.lua
index 3da898d..5427553 100644
--- a/resources/process-openmaptiles.lua
+++ b/resources/process-openmaptiles.lua
@@ -88,7 +88,7 @@ function node_function(node)

        -- Write 'poi'
        local rank, class, subclass = GetPOIRank(node)
-       if rank then WritePOI(node,class,subclass,rank) end
+       if rank then WritePOI(node,class,subclass,rank,true) end

        -- Write 'mountain_peak' and 'water_name'
        local natural = node:Find("natural")
@@ -402,7 +402,7 @@ function way_function(way)

        -- POIs ('poi' and 'poi_detail')
        local rank, class, subclass = GetPOIRank(way)
-       if rank then WritePOI(way,class,subclass,rank); return end
+       if rank then WritePOI(way,class,subclass,rank,false); return end

        -- Catch-all
        if (building~="" or write_name) and way:Holds("name") then
@@ -422,10 +422,14 @@ end
 -- Common functions

 -- Write a way centroid to POI layer
-function WritePOI(obj,class,subclass,rank)
+function WritePOI(obj,class,subclass,rank,isNode)
        local layer = "poi"
        if rank>4 then layer="poi_detail" end
-       obj:LayerAsCentroid(layer)
+       if isNode then
+               obj:Layer(layer, false)
+       else
+               obj:LayerAsCentroid(layer)
+       end
        SetNameAttributes(obj)
        obj:AttributeNumeric("rank", rank)
        obj:Attribute("class", class)
```

systemed:
Thanks for spotting! This should hopefully be fixed in https://github.com/systemed/tilemaker/commit/82d12335adcb1fbc26dc85bb58b861bc0858f69d (no profile change needed) - let me know how you get on.

toshelp:
I tried 82d1233 but didn't work.
I may have made some mistakes, but I made the following fixes.

```diff --git a/src/osm_lua_processing.cpp b/src/osm_lua_processing.cpp
index 26e1c3e..36522d5 100644
--- a/src/osm_lua_processing.cpp
+++ b/src/osm_lua_processing.cpp
@@ -291,7 +291,7 @@ void OsmLuaProcessing::LayerAsCentroid(const string &layerName) {
[...]

PR #201, opened 2021-03-06
Fix processing order of ways

systemed:
In 0fb2c00589d554a25c041bf30b9d369fcc7270c8 we accidentally broke the processing order for ways.

We need the way-node list available throughout setWay, particularly including the call to Lua's `way_function`. This means we can access (and correct) the geometry from the `way:Layer` function, check whether the way is closed, and so on.

Because we were only creating nodeVec _after_ `way_function` was called, we were actually using last time's cached linestring - so if that was in any way different from last time (e.g. last way was closed, this one isn't) then the results wouldn't be correct. This basically showed through in lots of output geometries going AWOL.

This PR fixes the order by creating nodeVec the first time we try to access a geometry, which might be in the `way:Layer`, `way:Length`, `way:Layer` or `way:LayerAsCentroid` calls.

(I've also included the macOS build fixes.)

@kleunen Does this look good to you?

### DIFF
--- a/include/osm_lua_processing.h
+++ b/include/osm_lua_processing.h
@@ -177,10 +177,11 @@ private:
 	uint64_t osmID;							///< ID of OSM object (relations have decrementing way IDs)
 	int64_t originalOsmID;					///< Original OSM object ID
 	WayID newWayID = MAX_WAY_ID;			///< Decrementing new ID for relations
-	bool isWay, isRelation;					///< Way, node, relation?
+	bool isWay, isRelation, isClosed;		///< Way, node, relation?
 
 	int32_t lon1,latp1,lon2,latp2;			///< Start/end co-ordinates of OSM object
 	OSMStore::handle_t nodeVec;				///< node vector
+	NodeVec *nodeVecPtr;
   	WayVec *outerWayVec, *innerWayVec;      ///< way vectors
 
 	Linestring linestringCache;
--- a/include/osm_store.h
+++ b/include/osm_store.h
@@ -293,7 +293,7 @@ public:
 			throw std::out_of_range("Failed to store node " + std::to_string(i) + ", index out of range");
 		}
 
-		mLatpLons->resize(std::max(i + 1, mLatpLons->size()));
+		mLatpLons->resize(std::max<size_t>(i + 1, mLatpLons->size()));
 		(*mLatpLons)[i] = coord;
 	}
 
@@ -663,7 +663,7 @@ public:
 	handle_t store_linestring(generated &store, Input const &src)
 	{
 		perform_mmap_operation([&]() {
-			store.linestring_store->resize(store.linestring_store->size() + 1);
+			store.linestring_store->emplace_back();
 			boost::geometry::assign(store.linestring_store->back(), src);
 		});
 
--- a/include/osm_store.h
+++ b/include/osm_store.h
@@ -628,11 +628,6 @@ public:
 		return ls;
 	}
 
-	bool wayIsClosed(handle_t handle) {
-		auto const &way = retrieve<WayStore::nodeid_vector_t>(handle);
-		return way.empty() || (way.front() == way.back());
-	}
-
 	handle_t relations_insert_front(WayID i, const WayVec &outerWayVec, const WayVec &innerWayVec) {
 		handle_t result;
 		perform_mmap_operation([&]() {
--- a/src/osm_lua_processing.cpp
+++ b/src/osm_lua_processing.cpp
@@ -132,11 +132,8 @@ bool OsmLuaProcessing::Intersects(const string &layerName) {
 // Returns whether it is closed polygon
 bool OsmLuaProcessing::IsClosed() const {
 	if (!isWay) return false; // nonsense: it isn't a way
-	if (isRelation) {
-		return true; // TODO: check it when non-multipolygon are supported
-	} else {
-		return osmStore.wayIsClosed(nodeVec);
-	}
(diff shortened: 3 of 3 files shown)


kleunen:
Yes, but then the 'wayIsClosed' in osm_store can be removed I guess, because it is not used anywhere else.

include/osm_lua_processing.h
```diff
@@ -177,10 +177,11 @@ class OsmLuaProcessing : public PbfReaderOutput {
 	uint64_t osmID;							///< ID of OSM object (relations have decrementing way IDs)
 	int64_t originalOsmID;					///< Original OSM object ID
 	WayID newWayID = MAX_WAY_ID;			///< Decrementing new ID for relations
-	bool isWay, isRelation;					///< Way, node, relation?
+	bool isWay, isRelation, isClosed;		///< Way, node, relation?
 
 	int32_t lon1,latp1,lon2,latp2;			///< Start/end co-ordinates of OSM object
 	OSMStore::handle_t nodeVec;				///< node vector
+	NodeVec *nodeVecPtr;
```
kleunen, line 184: You can actually obtain the lasted stored nodeVec also from the osmStore, using the nodeVec handle. But I think that it not actually needed. 

TBH i do not completely follow the flow of handle these different points/linestring/polygons, but I guess it makes it confusing that they are called from the to-be-defined lua script. 

But if you can use the nodeVecPtr instead of the handle, i guess that is prefered. And possibly, you can remove the handle all together ?

systemed: Yes, we could probably do a bit of tidying here. I'll merge this for now just to fix the issue, but then look at tidying it later.

kleunen: I was not really sure if the nodeVecPtr stayed valid when other functions were called. So that is why I was using the handle to the last stored linestring in osm_store to call certain function:
````
https://github.com/kleunen/tilemaker/blob/master/src/osm_lua_processing.cpp#L204
````

But if you are sure nodeVecPtr stays valid, you can use that one also :)

kleunen:
The CI is now broking, it is not resizing the mmap file anymore, it just keeps saying: 
Resizing osm store to size: 0M                
Resizing osm store to size: 0M                
Resizing osm store to size: 0M                
Resizing osm store to size: 0M                
Resizing osm store to size: 0M                
Resizing osm store to size: 0M                
Resizing osm store to size: 0M                
Resizing osm store to size: 0M                
Resizing osm store to size: 0M                
Resizing osm store to size: 0M                
Resizing osm store to size: 0M                
Resizing osm store to size: 0M

kleunen:
> Because we were only creating nodeVec _after_ `way_function` was called, we were actually using last time's cached linestring - so if that was in any way different from last time (e.g. last way was closed, this one isn't) then the results wouldn't be correct. This basically showed through in lots of output geometries going AWOL.

But then how do you access the nodeVec, before way_function is called ? Because only in way_function, the pointers to the nodeVec are set ?

kleunen:
oh, i understand. way_function is called from setWay. But can you not just store and create nodeVec before calling way_function ?

systemed:
We could, but we don't want to push it into the waystore before running `way_function` - if we're not actually going to use the way, we don't want to store it in memory. But there's definitely some tidying that could be done here. :)

Not sure what's going on with the CI - looking at that.

kleunen:
I think some pointer is now being dereferenced without being initialized. I think you can better just remove the nodeVec handle and change the functions in osmStore back to using begin/end iterator as it was before:

````
    template<class WayIt>
    Polygon nodeListPolygon(WayIt begin, WayIt end) {
        Polygon poly;
        fillPoints(poly.outer(), begin, end);
        geom::correct(poly);
        return poly;
    }

    // Way -> Linestring
    template<class WayIt>
    Linestring nodeListLinestring(WayIt begin, WayIt end) const {
        Linestring ls;
        fillPoints(ls, begin, end);
          return ls;
    }
````

And in osm_lua_processing.cpp: 
````
linestringCache = osmStore.nodeListLinestring(nodeVecPtr->cbegin(),nodeVecPtr->cend());
polygonCache = osmStore.nodeListPolygon(nodeVecPtr->cbegin(), nodeVecPtr->cend());

````

systemed:
That works better - thanks. `osmStore.ways_insert_back(osmID, *nodeVecPtr)` wasn't too clever when osmID was a 2^64-- ID for a way generated from a relation! Exasperatingly the problem wasn't showing up on the Northamptonshire extract I was testing with, but I've just tested this against a larger extract (Scotland) and it works now.

kleunen:
Maybe remove the previous commits from the git repository. Do you know how to do this? Reset the head and perform force push

kleunen:
Did we make some kind of other mistake ? Because I have no city names / street names on the maps anymore. Or am I doing something wrong now?

kleunen:
ah no, it was really a mistake in my html/style probably. I will continue conversion of Europe :)

systemed:
Whew! Yes, working fine for me too.

There's an issue with motorway_link ways being missing but that's a problem with the Lua script and I know what it is - will push a change later.

kleunen:
ah ok. Maybe I should wait for some more fixes to the lua script. I am actually generating the full of Europe now, using a makefile and the extracts of geofabrik. It is actually going suprisingly well now on this 8G machine. It already generated France and Germany and many smaller countries. It was no problem actually.

systemed:
I think the motorway_link issue and #200 are the only ones I know about (and #200 can be easily solved by supporting LayerAsCentroid for nodes; I'll add that later).

systemed:
Both of those should be fixed now.

PR #202, opened 2021-03-07
Perform static build of tilemaker

kleunen:
Build a static executable with CI on ubuntu 16.04. Allows the user to download and run the executable without installing any dependencies. You can download the executable from the CI artifacts and provide it on the website as release.


systemed:
Thanks!

PR #205, opened 2021-03-12
Allow generation of pbf index file

kleunen:
The following branch allow generating an "index" file of a pbf file. This allows multiple generators or subsequent runs of tilemaker to use the index file instead of loading. 

Create an index file from an input file:
````console
./tilemaker --index --compact --input overijssel-latest.osm.pbf --output overijssel-latest.mbtiles 
  --process process-openmaptiles.lua --config config-openmaptiles.json
````

The file 'overijssel-latest.osm.pbf.idx' is generated. Now use the index to generate the mbtiles:
````console
./tilemaker --compact --input overijssel-latest.osm.pbf --output overijssel-latest.mbtiles
  --process process-openmaptiles.lua --config config-openmaptiles.json
````

You can use both compact/non-compact. Just don't mix generating the index with compact and generating the mbtiles without compact. In the output it says:

````
Using index to generate tiles: overijssel-latest.osm.pbf.idx
Generate from index file
````

### DIFF
--- a/include/osm_lua_processing.h
+++ b/include/osm_lua_processing.h
@@ -41,7 +41,7 @@ public:
 	// ----	initialization routines
 
 	OsmLuaProcessing(
-        OSMStore &osmStore,
+        OSMStore const *indexStore, OSMStore &osmStore,
         const class Config &configIn, class LayerDefinition &layers, 
 		const std::string &luaFile,
 		const class ShpMemTiles &shpMemTiles, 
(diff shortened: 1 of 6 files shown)


systemed:
Works nicely - thank you!

kleunen:
Only thing is, it uses the "significant node keys" from the lua file to determine which nodes are stored. So if this changes in the lua file, you have to regenerate the index file. Maybe this is something to fix.

PR #209, opened 2021-03-21
Use unordered_set instead of sort to calculate unique output objects

kleunen:
Use unordered_set instead of sort to generate unique list of output objects per layer.


kleunen:
Not sure if it actually is faster, so i am performing a measurement now

kleunen:
I don't measure any speedup, probably because I use the renum files, they are already sorted ...

PR #210, opened 2021-03-21
Small optimization: store multipolygon in mmap in place

kleunen:
This is a small optimization of multipolygons in the mmap file. It uses boost::geometry::assign to copy the multipolygon now, gives a small improvement to performance. Also, it annoyed me that I could not assign the mmap multipolygon with boost geometry assign, and now it is possible.


systemed:
Thanks!

kleunen:
This actually breaks the windows build, it seems also on CI it is crashing now. So maybe better to revert this commit for now.

PR #211, opened 2021-03-23
Added a link to Wouter's blog item about "Generating self-hosted maps…

stephan75:
… using tilemaker"


systemed:
Looks good. I need to revise the docs thoroughly but this is a great start for now. (I also registered tilemaker.org the other day :) ).

PR #212, opened 2021-03-23
Write metadata.json file when generating directory output

kleunen:
Create also a 'metadata.json' file when writing pbf files to disk instead of mbtiles. 
To use it, the webserver needs to include the header Content-Encoding: gzip when serving pbf files with compression. 

````
location ~ /map/.+pbf$ {
            # gzip Encoding and MIME type type
            add_header  Content-Encoding  gzip;
            gzip off;
            types { application/x-protobuf pbf; }
}

### DIFF
--- a/include/shared_data.h
+++ b/include/shared_data.h
@@ -47,7 +47,8 @@ public:
 			const std::string &indexName,
 			const std::string &writeTo);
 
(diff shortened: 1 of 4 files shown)


systemed:
Thanks!

PR #218, opened 2021-04-05
Fix margin tiles artifacts

kleunen:


systemed:
The margins are basically there to implement https://blog.cyclemap.link/2020-01-25-tilebuffer/ - I wonder if there's some rounding going on that shouldn't be.

kleunen:
ah ok, so the margins are needed. Yes, it is clearly an artifact from the clipping. There is some rounding going on, which causes the two lines to not line up after clipping.

kleunen:
When the geometries are written, the coordinates are converted back from double -> int. So you have a line, you clip it, but it is actually clipped at two different points. This upper tile clips it at a certain point, the lower tile clips it at a certain point. When writing the clipped points to the different mbtiles data in WriteGeometryVisitor, you are converting these two points into int/int again. Due to rounding, this may cause the lines to not line up anymore. 

I think this is what you are seeing

PR #219, opened 2021-04-05
Optionally set minzoom to write attributes

### LINKED ISSUE #180: Minzoom for attributes
When setting attributes for a layer, these attributes are written out at every possible zoom level.

Very often, you don't actually need the attributes at the lower zoom levels. For example, you might render rivers at z6, but not actually render their names until z13/14. So the attributes are just bloating the vector tiles for z6-z12/13.

We could have an optional minzoom parameter for the `Attribute` method:

    node:Attribute("name", "River Towy", 14)

This would be stored as a single byte in the `AttributePair` class. Because we now have a global output dictionary, it shouldn't bloat the object store too much. We'd then use this when writing out each object to tiles.

systemed:
This fixes #180 by providing a third, optional parameter for `object:Attribute` to specify the minimum zoom at which to write the attribute.

The use case is for layers such as roads and rivers, where at high zooms you want to show them as a line with a name label, but at low zooms you just want to show them as lines. Therefore there's no reason to write the names into the vector tiles at low zooms. For a typical semi-urban z11 tile on a style I'm working on, this reduces the tile size from 193kb to 169kb, which is a very useful saving.

There will be a small impact on memory usage by storing the minimum zoom as a single byte with each key/value pair, but nothing significant enough for `htop` to pick up - performance and memory usage were both on a par with current master.

@kleunen You might want to check that I haven't done anything silly to your `attribute_store` code!

### DIFF
--- a/include/attribute_store.h
+++ b/include/attribute_store.h
@@ -20,7 +20,7 @@
 
 struct AttributeStore
 {
-    using key_value_t = std::pair<std::string, vector_tile::Tile_Value>;    
+    using key_value_t = std::tuple<std::string, vector_tile::Tile_Value, char>;
 	enum class Index { BOOL, FLOAT, STRING };
 
 	static Index type_index(vector_tile::Tile_Value const &v)
@@ -57,7 +57,11 @@ struct AttributeStore
     
     struct key_value_less {
         bool operator()(key_value_t const &lhs, key_value_t const& rhs) const {            
-            return (lhs.first == rhs.first) ? compare(lhs.second, rhs.second) : (lhs.first < rhs.first);
+			return std::get<2>(lhs) != std::get<2>(rhs) ?
+				std::get<2>(lhs) < std::get<2>(rhs) :
+				(std::get<0>(lhs) == std::get<0>(rhs)) ?
+				compare(std::get<1>(lhs), std::get<1>(rhs)) :
+				(std::get<0>(lhs) < std::get<0>(rhs));
         }
     }; 
     
@@ -67,7 +71,11 @@ struct AttributeStore
         
     struct key_value_store_less {
         bool operator()(key_value_store_iter_t lhs, key_value_store_iter_t rhs) const {            
-            return (lhs->first == rhs->first) ? compare(lhs->second, rhs->second) : (lhs->first < rhs->first);
+			return std::get<2>(*lhs) != std::get<2>(*rhs) ?
+				std::get<2>(*lhs) < std::get<2>(*rhs) :
+				(std::get<0>(*lhs) == std::get<0>(*rhs)) ?
+				compare(std::get<1>(*lhs), std::get<1>(*rhs)) :
+				(std::get<0>(*lhs) < std::get<0>(*rhs));
         }
     }; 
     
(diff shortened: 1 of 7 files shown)


kleunen:
Maybe just reformat the less struct:
````C++
 struct key_value_store_less {
        bool operator()(key_value_store_iter_t lhs, key_value_store_iter_t rhs) const {            
			return 
				std::get<2>(*lhs) != std::get<2>(*rhs) ? std::get<2>(*lhs) < std::get<2>(*rhs) 
				: std::get<0>(*lhs) != std::get<0>(*rhs) ? std::get<0>(*lhs) < std::get<0>(*rhs)
				: compare(std::get<1>(*lhs), std::get<1>(*rhs)); 
        }
    }; 
````

File: include/attribute_store.h [hunk 1]
```diff
@@ -20,7 +20,7 @@
 
 struct AttributeStore
 {
-    using key_value_t = std::pair<std::string, vector_tile::Tile_Value>;    
+    using key_value_t = std::tuple<std::string, vector_tile::Tile_Value, char>;
```
kleunen, line 23: You can use a struct for this, instead of a tuple. That way you can name the fields, instead of std::get<0> std::get<1>

kleunen:
looks good now

src/osm_lua_processing.cpp
```diff
@@ -338,28 +338,31 @@ void OsmLuaProcessing::LayerAsCentroid(const string &layerName) {
...
 	vector_tile::Tile_Value v;
 	v.set_float_value(val);
-	outputs.back().second.push_back(attributeStore.store_key_value(key, v));
+	outputs.back().second.push_back(attributeStore.store_key_value(key, v, 0));
 	setVectorLayerMetadata(outputs.back().first->layer, key, 1);
 }
 
-void OsmLuaProcessing::AttributeBoolean(const string &key, const bool val) {
+void OsmLuaProcessing::AttributeBoolean(const string &key, const bool val) { AttributeBooleanWithMinZoom(key,val,0); }
+void OsmLuaProcessing::AttributeBooleanWithMinZoom(const string &key, const bool val, const char minzoom) {
```
kleunen, line 361: make minzoom = 0 (default value)

src/osm_lua_processing.cpp
```diff
@@ -338,28 +338,31 @@ void OsmLuaProcessing::LayerAsCentroid(const string &layerName) {
...
 	if (outputs.size()==0) { cerr << "Can't add Attribute " << key << " if no Layer set" << endl; return; }
 	vector_tile::Tile_Value v;
 	v.set_bool_value(val);
-	outputs.back().second.push_back(attributeStore.store_key_value(key, v));
+	outputs.back().second.push_back(attributeStore.store_key_value(key, v, 0));
```
kleunen, line 365: remove (, 0) when adding default

systemed: Argh, that should have been `, minzoom` anyway! Copy & paste error. Fixed.

The reason for having separate `Attribute` and `AttributeWithMinZoom` methods (rather than just one with default parameters) is to make the Kaguya Lua bindings work - it's not beautifully elegant but I think it's better than using KAGUYA_FUNCTION_OVERLOADS, which is the other way that Kaguya provides.

kleunen:
You can also use a squash and merge when merging to the main branch:
https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges

That way you can remove those small "fix" commits. 
Not sure if you ever used this.

systemed:
Done - thanks!

PR #221, opened 2021-04-09
Combine linestrings only at specified zoom levels

systemed:
tilemaker has long had the ability to combine adjacent geometries, usually invoked with the `--combine` flag. The code to combine polygons was very slow, making it impractical for most tile generation. However, the code to combine linestrings is fast, and accounts for the vast majority of the tilesize saving.

This PR restricts the combine behaviour to look at linestrings only. It is now controlled by a `combine_below` parameter in the `settings` section of the JSON config file. For example, `"combine_below": 13` will run the linestring combine code on z1-12.

Tilesize results (OpenMapTiles-compatible schema, zooming in to Oxford):

````
Zoom   Before   After   Latest
6      191kb    149kb   142kb
7      1079kb   455kb   265kb
8      345kb    169kb   121kb
9      154kb    84kb    64kb
10     98kb     72kb    64kb
11     140kb    103kb   96kb
12     120kb    89kb    83kb
13     95kb     72kb    69kb
14     199kb    181kb   [not tested]
````

.mbtiles generation time (Great Britain extract), before:

````
real	33m41.744s
user	233m29.177s
sys	1m36.812s
````

.mbtiles generation time (Great Britain extract), after:

````
real	37m18.094s
user	291m29.527s
sys	2m1.871s
````

So there's a 10% time penalty, but a massive tilesize improvement at certain zoom levels.

I've set `"combine_below": 14` in the OMT-compatible config file, so that by default we reduce tilesize at the levels where it makes the most difference.

Bonus is that this completely removes the Clipper dependency, so we have a simpler codebase and faster compile times!

### CODE REVIEW

**Comment by systemed** [comment]
Further commit also joins linestrings within a MultiLinestring, as otherwise we tend to get lots of output like [AB, CD, BC]. This shrinks tilesizes by up to 20% at z9.

**Comment by kleunen** [comment]
That is very good. I thought this combine would never a good performance / filesize reduction tradeoff. But this looks good

**Comment by systemed** [comment]
Just tweaked it slightly to also fix the [DEFG, ABCD] case. I'll run another test on the GB extract and then this should be ready for merging, I think.

With this and a bit of simplification work I've got my WIP z11 tiles down from 369kb to 55kb 😃

**Comment by kleunen** [comment]
You could try a boost flat_map for the start and endpoints. Possibly this improves performance. Because it requires less allocation. 

And maybe prereserve the number of entries and maybe reuse the map between function calls. Althought it is not a very clean solution.

File: src/tile_worker.cpp [hunk 1]
```diff
@@ -7,6 +7,41 @@
...
+			Point lastPoint = ls[ls.size()-1];
+			auto foundStart = startPoints.find(xy_pair(lastPoint.x(),lastPoint.y()));
+			if (foundStart == startPoints.end()) break;
+			unsigned idx = foundStart->second;
+			if (added[idx]) break;
+			for (unsigned j=1; j<input[idx].size(); j++) ls.emplace_back(input[idx][j]);
+			added[idx] = true;
+		}
+		output.resize(output.size()+1);
+		output[output.size()-1] = std::move(ls);
```
kleunen, line 59: boost::geometry::append ?

systemed: Segfaults whenever I try it, oddly (even when retaining the `resize`, given that [the docs](https://www.boost.org/doc/libs/1_65_1/libs/geometry/doc/html/geometry/reference/algorithms/append.html) say "the MLS is not resized automatically").

systemed:
I've updated the table above with tilesizes for this version ("latest"). Generally better, but again z7-9 show the biggest improvements.

With `"combine_below": 14` on the GB extract:

````
real	39m10.940s
user	335m6.155s
sys	2m1.450s
````

I tried it with a flat_map but it was actually quite a lot slower - 46m59. I'm sure there are some further optimisations we could make but I'll merge this for now!

kleunen:
But I think the impact of the combining on the runtime is quite significant, you went from:
user	233m29.177s
to:
user	291m29.527s
to:
user	335m6.155s

You do not see the impact that much, because you have many cores available. If you would run this on a single core, the runtime would be 50% longer (233 -> 335). Because you have 12 cores?, you only see (335 - 233) / 12 = 9 minutes increase in runtime. Actually you only see a 7 minute increase, but this is only because you run on this many cores. 

I am running now on 4 cores, and I do see that the generation of the lower zoom levels takes significantly longer.

kleunen:
With "combine_below": 14,
Filled the tileset with good things at overijssel.mbtiles
real    3m0.468s
user    5m43.195s
sys     0m5.328s
54M Apr 11 08:34 overijssel.mbtiles

With "combine_below": 0,
Filled the tileset with good things at overijssel.mbtiles
real    2m21.191s
user    3m7.820s
sys     0m4.956s
56M Apr 11 08:37 overijssel.mbtiles

I guess some other stuff in the config influences as well, because before I had real time of ~1m50. Maybe the landcover, etc ..

systemed:
> So I would prefer having the --combine back :)

We could do a --no-combine option to override the config.json setting, maybe? I think most people will probably want the smaller tiles for production use (certainly given the big saving at z7-9), but for prototyping then quick generation is probably more important.

The combine algorithm is pretty naive - there may be ways to improve the algorithm, or optimise the implementation. I wonder if the `insert`s in `ReorderMultiLinestring` are where much of the slowdown is - particularly when inserting at the front of the linestring.

kleunen:
Yes, maybe a --no-combine option. But it depends on what the gain is in practice, on the 'overijssel' extract I only saw a 3/4% decrease in file size. It is possible only certain regions benefit the most ? depending on the openstreetmap data locally ?

systemed:
Because the whole .mbtiles is dominated by z14 tiles, then you don't see much of a difference in overall filesize - it's the lower zoom tiles that have the biggest reduction, and there aren't so many of these. (But, of course, in most applications they'll be more frequently viewed by the user than any given z14 tile would be!)

If you use a recentish version of Mapbox GL JS:

    <script src='https://api.mapbox.com/mapbox-gl-js/v1.10.1/mapbox-gl.js'></script>
    <link href='https://api.mapbox.com/mapbox-gl-js/v1.10.1/mapbox-gl.css' rel='stylesheet' />

and show the debug tile boundaries:

    map.showTileBoundaries=true;

then it renders the tilesize at the top left of each tile - quite interesting reading!

kleunen:
I see, yes, then there is a benefit, because most of the time you are rendering the lower zoom levels. So it would benefit the user if they are downloaded and rendered more quickly.

systemed:
Perhaps interesting: Boost has a 'connect' extension that does largely the same thing. I couldn't get it to build though (the extensions are largely not maintained and tend to break against newer Boost versions, which is a shame because 'dissolve' would be really useful for us). https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/extensions/algorithms/connect.hpp

kleunen:
I guess no one is actively maintaining these extensions, someone wrote it in the past, but current priorities are set elsewhere. It would be good if several openstreetmap projects combine there effort and focus on using the same set of libraries. But if I look, i guess this is what is happening, because I see the same libraries popping up in different OSM projects.

systemed:
I've found an optimisation to this which I'll merge after #222.

kleunen:
I guess if it is only useful for debugging to disable. I can just edit the config file.

kleunen:
For your combining algorithm, is the sorting of the output objects still required?

systemed:
It expects objects of the same geometry type (i.e. linestring) and attributes to be sequential:

https://github.com/systemed/tilemaker/blob/3eb7c716f6a6fb30263631cc8b40581338b3b477/src/tile_worker.cpp#L74-L76

freeExec:
It is a pity that the combination of adjacent polygons will be removed. I use it to combine forests and buildings. This allows them to be displayed without unnecessary seams.

systemed:
Do you have an example I could look at (i.e. a particular building/forest that splits across zoom levels)?

We could in theory support merging across polygons by running Boost's `union_` across them, as we do for linestrings (but without the reordering step afterwards). Previously we were using Clipper, which was very slow, but `union_` is more robust in recent versions of Boost.

freeExec:
The seams are not like artifacts, but like the outline of the perimeter of the building.
Here is an example where the buildings are drawn together and we can see the outline of each building. If you combine them, then only the outer contour will be visible, without seams on the inside.

https://www.openstreetmap.org/#map=19/53.24265/34.37348

systemed:
Great, thank you. I'll benchmark using `union_` for polygons.

systemed:
Ok; we can definitely do this. Here's an example with `union_`

Screenshot 2021-04-12 at 21 01 06

vs the original in OSM:

Screenshot 2021-04-12 at 21 01 17

It's certainly slow but I don't think it's slower than the old Clipper-based code. I wouldn't want it turned on by default but we can offer it as an option. My first thought is that we could do it as a per-layer option, so you could specify it individually for (say) forest and building layers, without having to slow down coastlines etc. - would that work for you?

systemed:
Polygon-merging WIP at https://github.com/systemed/tilemaker/tree/merge_polygons - you'll need to set the `"combine_below": 15` (or whatever) in the JSON config.

systemed:
Ok - I've updated the branch. Polygon merging is disabled by default, but you can set it per-layer by setting combine_polygons_below in that layer (for example, `"combine_polygons_below": 15`).

PR #222, opened 2021-04-11
Cleanup to tile coordinates generation

kleunen:
This PR cleans up the tile coordinate generation and moves the obtaining of the OutputObject lists for a tile in a worker. This reduces the memory requirement, because this list of OutputObjects for a tile is now generated once the tile is about the be generated instead of beforehand. 

Finally, a patch for closing the mmap file before resizing, this is required on windows to resize the mmap file.

### DIFF
--- a/src/tilemaker.cpp
+++ b/src/tilemaker.cpp
@@ -410,17 +410,7 @@ int main(int argc, char* argv[]) {
 	// ----	Initialise SharedData
 	std::vector<class TileDataSource *> sources = {&osmMemTiles, &shpMemTiles};
 
-	std::map<uint, TileData> tileData;
-	std::size_t total_tiles = 0;
-
-	for (uint zoom=config.startZoom; zoom<=config.endZoom; zoom++) {
-		tileData.emplace(std::piecewise_construct,
-				std::forward_as_tuple(zoom), 
-				std::forward_as_tuple(sources, zoom));
-		total_tiles += tileData.at(zoom).GetTilesAtZoomSize();
-	}
-
-	class SharedData sharedData(config, layers, tileData);
+	class SharedData sharedData(config, layers);
 	sharedData.outputFile = outputFile;
 	sharedData.sqlite = sqlite;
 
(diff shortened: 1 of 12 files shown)


src/tilemaker.cpp
```diff
@@ -410,17 +410,17 @@ int main(int argc, char* argv[]) {
...
+	/* std::map<uint, TileData> tileData;
 	std::size_t total_tiles = 0;
 
 	for (uint zoom=config.startZoom; zoom<=config.endZoom; zoom++) {
 		tileData.emplace(std::piecewise_construct,
 				std::forward_as_tuple(zoom), 
 				std::forward_as_tuple(sources, zoom));
 		total_tiles += tileData.at(zoom).GetTilesAtZoomSize();
-	}
+	} */
```
kleunen, line 424: just remove

kleunen:
I see my windows version crashing now during the conversion. No error, it just terminates. Not sure if this is the linestring merging or this commit.

kleunen:
It does work on linux. The map feels faster with the linestring merging, no scientific proof for this though :D

systemed:
I wouldn't have thought the linestring merging would make a difference on Windows - you could test it by running with `"combine_below": 0` (or just removing it).

With the post-222 optimisation the GB extract will be

````
real	34m22.158s
user	300m50.460s
sys	1m51.753s
````

i.e. user time is roughly halfway between pre-merging and current, real is nearly back to pre-merging (on a 16-core machine).

kleunen:
> I wouldn't have thought the linestring merging would make a difference on Windows - you could test it by running with `"combine_below": 0` (or just removing it).

Yes, i tried this. But it did not help, so it was not the merging. I also had some changes to the workers, to convert 25 tiles per worker instead of one. Maybe this was the problem, i will try now the executable from this commit.

kleunen:
> i.e. user time is roughly halfway between pre-merging and current, real is nearly back to pre-merging (on a 16-core machine).

So it is quite a significant improvement ?

systemed:
Yes - basically in CheckNextObjectAndMerge, it moves the call to ReorderMultilineString outside the while loop, so we only call it once for each union of linestrings.

kleunen:
The crash had to do with somethings I changed to improve the multicore performance, that was not part of this PR. Seems to run OK on windows also with these fixes and with merging

systemed:
Great - the code looks a fair bit cleaner with this, it works well on my machine and memory usage does seem to be less. Thanks!

PR #224, opened 2021-04-17
Don't generate tiles outside bounding box of map

kleunen:
Following commit checks if the to-be-generated tile has overlap with the bounding box of the map. If the tile is outside the bounding box, it won't be generated. This allows exporting a map exactly on the lines of the tiles for a certain zoom level, and then using --merge to stitch the maps together. 

Only bounds in metadata are not updated when using --merge.


systemed:
This seems to work fine for me post-#225 - still ok to merge?

kleunen:
Yes it is ok to merge

systemed:
Thanks!

Issue #192, opened 2020-12-27
--merge is misleading


uvwxy:
>  --merge                      merge with existing .mbtiles (overwrites otherwise)

it should rather be

>  --merge                      overwrite tiles in existing .mbtiles

because the resulting output is not as expected.

If I run tilemaker on two .pbf files with `--merge` my expectation was that the results from both files are usable/viewable, and not just the first + overwritten by the second.

I assume this is not a trivial issue.

kleunen:
I have noticed this issue too. I set the boundingbox on the edge of a tile, but tilemaker generates tiles also outside the boundingbox. I have made an attempt to fix this, but have not been able too fix it

systemed:
This should be fixed now in #225 and #224.

PR #225, opened 2021-04-17
Merge tile contents

systemed:
This addresses the other half of #192 by merging data within tiles. If (for example) you create an .mbtiles of Belgium:

    tilemaker --input belgium.osm.pbf --output map.mbtiles

you can then run this to add the Netherlands:

    tilemaker --input netherlands.osm.pbf --output map.mbtiles --merge

and all shared tiles will be merged, rather than overwritten.

**Blocking issue:**

This is not thread-safe and will crash unless you pass `--threads 1`. The issue is `decompress_string` in helpers.cpp. As it stands `readTileAndUncompress`(mbtiles.cpp) is pretty ugly - it reads from the .mbtiles as a `vector<char>` (if we try a string, then a 0 is taken as end-of-string); then copies to a `std::string`; and then decompresses that, so that we can do `tile.ParseFromString`. 

It would probably be better to have a thread-safe version of `decompress_string` that can work directly on a `vector<char>`, or something. My brain is a bit frazzled and I haven't worked out how to do this.

**Non-blocking issues:**

- `readTileAndUncompress` runs two SQLite queries; first to check whether a tile exists, then to read it. In theory we could just do this with the second query, but `sqlite_modern_cpp.h` throws a runtime_error and for some reason I haven't been able to catch it.
- Currently we assume the .mbtiles contains compressed tiles. It would probably be better to take this from `sharedData.config.compress`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
 
     - name: Build dependencies
       run: |
-        vcpkg install --triplet=x64-windows-static-md lua shapelib zlib protobuf[zlib] sqlite3 boost-program-options boost-filesystem boost-geometry boost-system boost-asio boost-interprocess 
+        vcpkg install --triplet=x64-windows-static-md lua shapelib zlib protobuf[zlib] sqlite3 boost-program-options boost-filesystem boost-geometry boost-system boost-asio boost-interprocess boost-iostreams
 
     - name: Build tilemaker
       run: |
@@ -77,7 +77,7 @@ jobs:
 
     - name: Build dependencies
       run: |
-        vcpkg install --triplet=${{ matrix.triplet }} lua shapelib zlib protobuf[zlib] sqlite3 boost-program-options boost-filesystem boost-geometry boost-system boost-asio boost-interprocess 
+        vcpkg install --triplet=${{ matrix.triplet }} lua shapelib zlib protobuf[zlib] sqlite3 boost-program-options boost-filesystem boost-geometry boost-system boost-asio boost-interprocess boost-iostreams
 
     - name: Build tilemaker
       run: |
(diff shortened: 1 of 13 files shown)


kleunen:
Instead of: 
````
std::string compress_string(const std::string& str,
                            int compressionlevel,
                            bool asGzip) {
````

Use: 
````
std::string compress_string(char const *str, std::size_t str_len,
                            int compressionlevel,
                            bool asGzip) {
``

That way you can use both std::string as well as std::vector<char>

kleunen:
For split_string in helpers.h, you can use boost actually: 
https://www.geeksforgeeks.org/boostsplit-c-library/

systemed:
That's neater - thanks! I've just moved this bit of decompression for now. And I found the thread issue was actually being caused by not locking around the first sqlite query. Something's now failing with the build, however.

kleunen:
> That's neater - thanks! I've just moved this bit of decompression for now. And I found the thread issue was actually being caused by not locking around the first sqlite query. Something's now failing with the build, however.

But you add boost-iostreams to the github CI vcpkg install (.github/workflows/ci.yml)
````
 vcpkg install --triplet=x64-windows-static-md lua shapelib zlib protobuf[zlib] sqlite3 boost-program-options boost-filesystem boost-geometry boost-system boost-asio boost-interprocess boost-iostreams
````

Because CI is not building now

kleunen:
> That's neater - thanks! I've just moved this bit of decompression for now. And I found the thread issue was actually being caused by not locking around the first sqlite query. Something's now failing with the build, however.

Yes, I was wondering what the thread issue was. Had a look at the decompression code and was like: hmm, does seem thread-safe :D

kleunen:
But is important the "bounds" metadata is also updated, i have issues with mapbox-gl if there are no bounds set or incorrect bounds. So select min/max x/y from the tiles in the database, and base the bounds of the map on this.

systemed:
Ok - it now enlarges the bounding box to include both the new and previous extents.

kleunen:
Try also merge with #224

kleunen:
I think if you specify --merge and the mbtiles file does not exist, you get a segfault.

PR #226, opened 2021-04-19
Tile generation use blocks of 100 tiles per worker

kleunen:
Generate blocks of 100 tiles per thread pool worker. This reduces memory consumption a bit on large areas and improves the tile progress output.

### DIFF
--- a/src/tilemaker.cpp
+++ b/src/tilemaker.cpp
@@ -487,54 +487,45 @@ int main(int argc, char* argv[]) {
 		std::mutex io_mutex;
 
 		// Loop through tiles
-		uint tc = 0;
+		std::size_t tc = 0;
 
-		std::size_t total_tiles = 0;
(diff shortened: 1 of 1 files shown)


systemed:
Looks good! I haven't timed it yet but it seems a little faster.

I get slightly weird progress output at the end:

    Zoom level 4256283078, writing tile 1890 of 1990               

followed by

    Zoom level 12, writing tile 1990 of 1990

which seems odd as it should finish at z14 (not 4256283078 :) ).

kleunen:
That is strange

kleunen:
It is possible the zoom level sometimes jumps back. Because tiles from higher zoom level may finish later. But the value you mentioned looks corrupted

kleunen:
It was accessing end_index, while last element is (end_index - 1)

systemed:
That seems happier! Thank you.

PR #232, opened 2021-04-30
Add ferry routes to output tiles

QuentinC:
Hello,

Here is a simple pull-request that adds ferry routes to the tiles.
Here is the OSM corresponding doc: https://wiki.openstreetmap.org/wiki/Tag:route%3Dferry

Regards !


systemed:
That's great - thank you!

PR #239, opened 2021-05-09
Non self-intersecting simplify

kleunen:
Simplify algorithm which uses an rtree to lookup if an intersection occurs when creating a new simplified edge. Is uses the approach from:
[Simplify without intersection](https://www.jasondavies.com/simplify/)


systemed:
Wow, looks like an interesting approach - thank you. I'll have a play tomorrow.

kleunen:
@systemd Try the simplifier now in the current state. 
Run it with '--verbose'. 

It is quite solid now, i think still under some rare conditions problems may occur. I see now almost only polygons which have becomes self-intersecting by the clipping operation.

systemed:
Ok, running a test now...

kleunen:
I do see the following in my Netherlands extract: ijsselmeer is getting filled up.  not sure if this is because of the simplification
image

systemed:
Obligatory Great Britain test:

````
real	41m2.625s
user	323m7.198s
sys	1m39.056s
````

A bit slower but most of this seemed to be one difficult z7 tile which was the last to complete (even after z14).

Actual tile rendering seemed much faster than the slow rendering I saw last time. Filesize is smaller too (1.18GB vs 1.22GB).

At z12 a few islands seemed to go missing - see for example St Michael's Mount and Staffa:

Screenshot 2021-05-12 at 14 09 00

Screenshot 2021-05-12 at 14 09 42

and this bay on Taransay disappears north of the tile border (i.e. up to Uidh) at z11/z12, but is present at z10 and z13 - I wonder if that's similar to the Ijsselmeer issue?:

Screenshot 2021-05-12 at 14 12 55

**Comment by kleunen** [comment]
Yes, i am not sure if this was the case in the old simplifier (boost simplifier). We should actually compare the output of this simplifier with the old simplifier. Do you know if that island was also missing in the old renderer ? 

With this scotland extract, i think the islands are missing due to self-intersection in the polygons. I have a small extract of these islands and there the self-intersection gets resolved by the simplifier and then the island is visible again .. While in a large extract it does not resolve, very strange behaviour.

**Comment by kleunen** [comment]
There is still some self-intersection errors occuring in the clipping. So this might also cause the output to break. You can see that during the generation. Some polygons they get clipped at higher zoom levels, they render OK. When you zoom out, they don't get clipped anymore and they have self-intersection. When you zoom out further, then sometimes the self-intersection gets resolved by the simplifier.

**Comment by kleunen** [comment]
image
vs
image

File: src/osm_lua_processing.cpp [hunk 1]
```diff
@@ -344,7 +344,7 @@ void OsmLuaProcessing::Layer(const string &layerName, bool area) {
 				mp.push_back(p);
 			}
 
-            CorrectGeometry(mp);
+            if(!CorrectGeometry(mp)) return;
```
kleunen, line 347: I remove now polygons on input if they have self-intersection. Possibly, this is not a good idea. They might get rendered after simplification ?

systemed:
Just checked with an old .mbtiles produced using the original Boost simplify and both St Michael's Mount and Staffa were rendered as you'd expect.

I'll try it without removing self-intersecting polygons.

kleunen:
> I'll try it without removing self-intersecting polygons

If it is different behaviour at different zoom levels, it is probably the simplifier. So if it was visible at Z10 and not Z11 and Z12, it is probably simplification. Have a look if the polygons are at tile boundary. And run tilemaker with --verbose, you can see if the simplifier is generating self-intersection in the polygons. 

It is possible the self-intersection occurs at tiling and the old simplifier simply was able to remove the self-intersection and the new simplifier is not.

kleunen:
Yes, with these small area such as this bay: 
image

After much simplification it is possible this complete area "collapses" and so vanishes. I also check if the distance between to points is within a certain distance, and combine them if they are. But maybe that is not a good idea.

src/write_geometry.cpp
```diff
@@ -1,10 +1,117 @@
...
+        auto start = nodes[entry];
+        auto middle = nodes[entry + 1];
+        auto end = nodes[entry + 2];                   
+                
+        simplify_segment line(input[start], input[end]);
+        double distance = 0.0;
+        for(auto i = start + 1; i < end; ++i) 
+            distance = std::max(distance, boost::geometry::distance(line, input[i]));          
+    
+        if((boost::geometry::distance(input[start], input[end]) < 2 * max_distance) || (distance < max_distance)) {
```
kleunen, line 51: This start-end distance check might not be a good idea ?

systemed:
> I'll try it without removing self-intersecting polygons.

Still missing even with the `return` commented out.

kleunen:
If I set simplifyLevel = 0, this staffa island still disappears! Only when I disable simplification completely, it is still there. Something odd going on there ...

systemed:
Removing this line

    if((boost::geometry::distance(input[start], input[end]) < 2 * max_distance) || (distance < max_distance)) {

results in some interesting geometries...!

Screenshot 2021-05-12 at 16 21 35

kleunen:
Yes, you should not remove it completely, because this means all edges will be removed :D

Change to:
````
if( (distance < max_distance)) {
````

But with this staffa island, i am also seeing some unexpected results.

systemed:
Yes - still the same issue with Staffa etc. if I just have `if( (distance < max_distance)) {`.

kleunen:
Yes, it has to do with this union between the inners that is performed. The union of two overlapping polygons, results in 2 polygons ? How is that possible.

kleunen:
I fixed combining the inners, this helps for the staffa island. You can try. Maybe need to do more testing and combine the polygons in the multipolygon. I will do this later.

kleunen:
I added the polygon combining. ijsselmeer is still missing at zoom level 5, i am not sure if it really related to the simplifier. Maybe it was like that already before

image

systemed:
Mostly there! Staffa, St Michael's Mount and that bay seem to be fixed.

North and South Uist (i.e. the southern half of the Outer Hebrides) are missing:

Screenshot 2021-05-13 at 10 17 11

Timing is:

````
real	46m48.498s
user	334m22.584s
sys	1m33.977s
````

kleunen:
These islands are filled now: 

image

kleunen:
Yes, the only way I can fix Uist, is by changing the round -> floor in coordinates. I pushed this now a minute ago, For me, this Scotland extract renders fine now. 

image

kleunen:
I still get these messages. Maybe resimplify with 2*simplifylevel if polygon has more than 65535 vertices ?
image

kleunen:
Looks so nice, you can print it and put it on the wall
image

kleunen:
> Timing is:
> 
> ```
> real	46m48.498s
> user	334m22.584s
> sys	1m33.977s
> ```

There is now also combining of polygons within a multipolygon being done, but I am not sure if this is needed. If a multipolygon with overlapping polygons valid ? I would say yes, in that case, the union of these polygons is not actually needed.

kleunen:
Hmm, if I look here: https://gis.stackexchange.com/questions/225368/understanding-difference-between-polygon-and-multipolygon-for-shapefiles-in-qgis/225373

2. The boundaries of any 2 Polygons that are elements of a MultiPolygon may not “cross” and may touch at only a finite number of Points;

So they may not overlap, they may "touch".

systemed:
Latest fixes North/South Uist at z7 but not at z6. :(

Screenshot 2021-05-13 at 13 16 48

I've converted the tile into GeoJSON and QGIS reports invalid polygons. I tried putting an extra `geom::correct(current);` after the simplify operations but this doesn't fix it, sadly.

kleunen:
Really weird. Maybe get the latest extract from geofabrik?
https://mqtt.kleunen.nl/map/scotland.html

I have also seen that my bbox extract of this island has issues. and full scotland extract does not. So is really strange issue

systemed:
That is weird. It shouldn't be an issue with the Geofabrik data - the water polygons are from the shapefiles (https://osmdata.openstreetmap.de/data/water-polygons.html).

kleunen:
Try maybe render upto zoom level 6 and once with no simplification and once with a lot of simplification. See if any errors are removed from the polygon.

Does qgis give an error on what is wrong with yhe polygon?

systemed:
It doesn't. I fed the GeoJSON into an online validator (geojsonlint.com) and that suggested the winding order was wrong - "Polygons and MultiPolygons should follow the right-hand rule". Here's the relevant bit of the file: https://www.systemed.net/temp/invalid.geojson

systemed:
QGIS has a "Fix geometries" function so I tried that. Here's an interesting bit:

Screenshot 2021-05-13 at 14 54 23

The black lines with red X vertices are the invalid original; the brown polygons are the corrected multipolygon (sea/lakes). You can see a handful of invalid geometries that are fixed, including an entire firth (Loch Sgiopoirt).

kleunen:
In an earlier commit, i tried rounding the coordinates before simplification in an earlier commit. If the simplifier is self-interesection aware, this could help in resolving these intersection. Also now this overlapping polygon union is working correctly, this might actually resolves some issues.

kleunen:
If I perform the rounding of coordinates before simplification, you see a lot of self-intersection occurs because of the rounding of the nodes:

input multipolygon = Result from clipping
output multipolygon = After rounding and simplification 

output multipolygon has self-intersections
input multipolygon valid
output multipolygon has self-intersections
input multipolygon valid
output multipolygon has self-intersections
input multipolygon valid

kleunen:
I pushed this change for pre-rounding the coordinates of the polygons. You can see that a lot more self-intersection occurs because of this. Also the rounding should be "self-intersection aware". If the rounding causes a self-intersection, try to round to another coordinate.

kleunen:
In the simplify loop the rtree of the line segments of the ring is available. Possibly we can do a final check on the ring to see if some segments still intersect and try to move the coordinates a bit if they intersect.

kleunen:
image

kleunen:
Every coordinate can be rounded in any of these directions:
image

kleunen:
Another way to resolve these self-intersections is finding the offending line segments and force simplification of these segments. Even if they are moving more than the max_distance. But this might not always resolve the intersection.

systemed:
Latest code seems to work and is faster too!

````
real	34m31.425s
user	312m9.757s
sys	1m42.785s
````

kleunen:
http://www.angusj.com/delphi/clipper/documentation/Docs/Overview/Rounding.htm

kleunen:
It is not a "perfect" solution, probably works because of the simplifier. Some cases might still fail. But it would be good if it helps in practice.

systemed:
Cool. I'll play around with it a bit more tomorrow, and then all things being equal, we can merge it - thank you!

kleunen:
Maybe cleanup the commit a bit and move the simplify to coordinates.cpp and rename this geometry. There is also the tile intersection function there

kleunen:
Or geomtypes.h?

kleunen:
I have some ideas how possibly the simplifier can be further tweaked to remove more self intersection. So dont merge yet

kleunen:
I tried simplifying before clipping, but this made things actually worse. I would like to get ijsselmeer rendered at zoom level 5, but don't know to achieve this just yet.

systemed:
Ijsselmeer is a tagging issue rather than a geometry issue (you'll be pleased to hear).

The coastline shapefiles don't include Ijsselmeer, so it's rendered from OSM data. In OSM it's tagged as a lake (`natural=water`, `water=lake`). Our Lua profile sees those tags and puts it in the `water` VT layer, which is defined in config-openmaptiles.json with minzoom 6.

So to get Ijsselmeer to render at z5:

- change the `water` layer in config-openmaptiles.json to have minzoom 5
- add z5 support to `SetMinZoomByArea` in the Lua profile, probably something like `if area>ZRES4^2 then way:MinZoom(5) elseif [...]`
- define `ZRES4=9783.94` at the top of the Lua profile

(You could probably do the same for z4 if you like!)

kleunen:
Aha! I will try. And yes, i would like z4 also :D

kleunen:
That is also why oosterschelde was rendered, and westerschelde/grevelingenmeer/haringvliet where not. Oosterschelde is an open connection to the North Sea, while westerschelde/haringvliet/gerevelingenmeer are not.

kleunen:
On Z5 it displays:
image

On Z4 it is still missing:
image

kleunen:
i also need to add ZRES3?

systemed:
Yes, I'd think so:

- change the water layer in config-openmaptiles.json to have minzoom 4
- add z4 support to SetMinZoomByArea in the Lua profile - if area>ZRES3^2 then way:MinZoom(4) elseif [...]
- define ZRES3=19567.88 at the top of the Lua profile

You could maybe put a trace statement in just to see what the area actually is:

    if way:Find("name")=="IJsselmeer" then print("Ijsselmeer area is "..way:Area()) end

kleunen:
I think you may actually need to go down to Z2, because Lake Ladoga is even bigger:
https://en.wikipedia.org/wiki/Lake_Ladoga

And oosterschelde is even rendered at Z3, so ijsselmeer should be visible as well:
image

systemed:
It might actually be better to use Natural Earth shapefiles for low zoom levels: https://www.naturalearthdata.com/downloads/50m-physical-vectors/50m-lakes-reservoirs/

They're pre-simplified (you can choose from 10m, 50m or 110m resolution, but I'd guess 50m is about right) so will save processing time and complications.

kleunen:
I am happy for now, zoom level 4 is enough for me. I think you can just use the tilemaker simplification. At these lower zoom levels there is not much else there on the map. Something like caspian sea is always visible, at any zoom level.

kleunen:
image

kleunen:
@systemd That's it for now, i guess :)

Should I also make a PR for the lake zoom levels ? or you will have a look at this yourself ?

kleunen:
Maybe good also to validate the existing know render issues here on github. If they do not render correctly, there are still some small improvements we can try.

toshelp:
It seems that there are cases where polygons get worse.
I also agree with the verification.
simplify_invalid

simplify_check

systemed:
Though that is an invalid polygon anyway (it has a 'spike' in it) - ideally it should be fixed in OSM.

> @systemd That's it for now, i guess :)
> Should I also make a PR for the lake zoom levels ? or you will have a look at this yourself ?

I'm happy to have a look. I'll try and do a final test and merge this tomorrow.

kleunen:
But why is the spike not removed ?
https://github.com/systemed/tilemaker/pull/239/commits/8c7c4ba49ee6d11822d760988777ee6584121cf0#diff-c08fc418ad2c73829e88bcc4588acd8cbf7d8ba8ff706dc72020ab2512684ab6R130-R131

systemed:
It's an unusual one - it's not a spike which backtracks onto the same vertex (ABA), but rather one where the vertices are aligned perfectly such that it forms a spike anyway.

There appears to be an issue with `remove_spikes` anyway: https://github.com/boostorg/geometry/issues/603

kleunen:
ah ok :)

I am getting the feeling boost geometry is a bit unmaintained.

kleunen:
GEOS would not have been a bad choice:
https://www.tsusiatsoftware.net/jts/files/jts_secrets_foss4g2007.pdf

Slide 12: Polygon cleaning using buffer(0)
Slide 13: TopologyPreservingSimplifier

kleunen:
Ah this buffer is also available in boost geometry:
https://www.boost.org/doc/libs/1_76_0/libs/geometry/doc/html/geometry/reference/algorithms/buffer/buffer_4.html

So it is possible to restore a self-intersecting polygon using this buffer with distance = 0

systemed:
I think we're almost there. Holy Island (off Anglesey, Wales) seems to still have a problem - fine at all zoom levels except 8:

Screenshot 2021-05-18 at 20 22 18

Screenshot 2021-05-18 at 20 22 30

kleunen:
Can you export that geometry again and see if the polygon has an issue?

systemed:
Converting it to GeoJSON and loading it into QGIS shows that the polygon was dropped entirely at z8:

Screenshot 2021-05-19 at 18 05 18

That's just showing polygons - if I show polylines as well then you can see where it should be:

Screenshot 2021-05-19 at 18 06 13

kleunen:
So it is not even written / generated at this zoom level ?

systemed:
Exactly - or more specifically, it's an island in the source shapefile (i.e. a hole/inner in the ocean multipolygon), and the inner isn't being written to the tile at z8 only. All other zoom levels seem to be fine.

kleunen:
Ah yes. This was what I was wondering, if this island is actually an inner. 

There can be two cases where an inner is removed: 
1. Its area < max_distance * max_distance
2. The outer is simplified in such a way, that the inner goes outside the outer

systemed:
That makes sense. It's definitely not 1 (there are smaller islands and it shows at z7) so I guess we might be dealing with a rounding issue causing 2 again?

kleunen:
Well, case 1 is also possible if the orientation of the polygon gets inverted. In that case the area is -area of the inner. 

I don't know about 2. It is possible, but the simplifier works by removing points and checking the new line does not intersect an inner our outer segment. This new line would have to "jump" over the whole island to get to the other side. 

I can remove the "inner-inside-outer" check and let boost check the geometries. If boost detects any inners outside the polygon, we have to make sure this does not happen.

kleunen:
Ah yes, rounding may be possible. If due to rounding the inner is not completely covered by the outer, it will be added.

kleunen:
Ah yes, the issue is this covered_by check

kleunen:
I just removed the check now completely

kleunen:
Yes i still see boost complaining about self intersecting polygons. And then when i plot them in excel. They look perfectly fine.

kleunen:
We can add spike removal to the simplifier. By keeping an rtree of unprocessed points to the ring and make a line directly to a point further on the ring if the point is very close. But this maybe something for another pr

systemed:
Yes, I think so!

All 16 cores of my machine are currently occupied running OSRM for cycle.travel in North America, but hopefully it'll finish some time tomorrow 😆

kleunen:
That is for cycle travel right? I did notice cycle travel is not using tilemaker maps, right? :)

systemed:
The website isn't (there's a _lot_ of Leaflet-based code in there that would take a long time to convert), but the mobile app will!

kleunen:
It would also be nice if geofabrik not only provides extracts of regions but also rendered pbf files. And maybe provide a planet render using torrent. Would be nice if something like this could be provided. Maybe google is willing to provide a cloud server for a weekly render of the planet. Or microsoft.

systemed:
I think we can go with this. Performance is great and it seems to work 99.9% of the time. North Uist still disappears at z4 but frankly at that zoom level, detail is limited anyway. Ok to merge?

kleunen:
Yes that is ok. The clipping and rounding can still cause self intersection. Possibly in some conditions this is still visible.

systemed:
Superb - thank you. Definitely an epic PR :)

Issue #240, opened 2021-05-19
Weird behavior with "coastlines" shp file


QuentinC:
Hello,

I've been trying to use tilemaker with the "coastlines-split" from here: [coastlines-split-4326.zip](https://osmdata.openstreetmap.de/download/coastlines-split-4326.zip)

As this seems related to https://github.com/systemed/tilemaker/pull/239 I've also tried with this branch, but with the same result.
I am not using simplification for now.

The result is not the expected one, some lines are missing, and some other are running straight through the land...

Level 6:
![6](https://user-images.githubusercontent.com/3394508/118885521-2cec0e80-b8f8-11eb-82c5-8aa2a9003991.jpg)
Level 7:
![7](https://user-images.githubusercontent.com/3394508/118885523-2d84a500-b8f8-11eb-97d7-672494ce2d85.jpg)
Level 8:
![8](https://user-images.githubusercontent.com/3394508/118885524-2d84a500-b8f8-11eb-94cc-88baeefc53da.jpg)
Level 9 seems fine:
![9](https://user-images.githubusercontent.com/3394508/118885525-2e1d3b80-b8f8-11eb-89d2-9da8f5f454f7.jpg)

Here is my basic config file:
```
{
	"layers": {
		"coastline":	     { "minzoom":  0, "maxzoom": 14},
		"coastline_hig":     { "minzoom":  6, "maxzoom": 14,"source": "data/res/coastlines.shp",  "source_columns": ["featurecla"], "write_to": "coastline" }
	},
	"settings": {
		"minzoom": 0,
		"maxzoom": 10,
		"basezoom": 10,
		"include_ids": false,
		"bounding_box": [-5, 46, 1, 50],
		"default_view": [-2.6, 48, 7],
		"combine_below": 14,
		"name": "Tilemaker to OpenMapTiles schema",
		"version": "3.0",
		"description": "Tile config based on OpenMapTiles schema",
		"compress": "gzip",
		"filemetadata": {
			"tilejson": "2.0.0", 
			"scheme": "xyz", 
			"type": "baselayer", 
			"format": "pbf"
		}
	}
}
```

Here is the resulting mbtiles file: [out.zip](https://github.com/systemed/tilemaker/files/6511619/out.zip)

I am doing anything wrong ?

Thanks !

kleunen:
Most likely this is in the input, can you open the shapefile in qgit ?

QuentinC:
I dont have any issue with this shape file in QGIS:

[...]

PR #243, opened 2021-05-26
Parallel loading of osm.pbf file

kleunen:
This PR performs parallel loading of the osm.pbf file. There is still a lot of locking going on, but it seems some level of parallel processing of the pbf file is possible. Have a try to see how it works. 

I removed the index building in this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
 
     - name: Build dependencies
       run: |
-        vcpkg install --triplet=x64-windows-static-md lua shapelib zlib protobuf[zlib] sqlite3 boost-program-options boost-filesystem boost-geometry boost-system boost-asio boost-interprocess boost-iostreams
+        vcpkg install --triplet=x64-windows-static-md lua shapelib zlib protobuf[zlib] sqlite3 boost-program-options boost-filesystem boost-geometry boost-system boost-asio boost-interprocess boost-iostreams boost-sort
 
     - name: Build tilemaker
       run: |
@@ -77,7 +77,7 @@ jobs:
 
     - name: Build dependencies
       run: |
-        vcpkg install --triplet=${{ matrix.triplet }} lua shapelib zlib protobuf[zlib] sqlite3 boost-program-options boost-filesystem boost-geometry boost-system boost-asio boost-interprocess boost-iostreams
+        vcpkg install --triplet=${{ matrix.triplet }} lua shapelib zlib protobuf[zlib] sqlite3 boost-program-options boost-filesystem boost-geometry boost-system boost-asio boost-interprocess boost-iostreams boost-sort
 
     - name: Build tilemaker
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,12 +32,12 @@ jobs:
       run: |
         mkdir ${{ github.workspace }}\build        
         cd ${{ github.workspace }}\build && cmake -DTILEMAKER_BUILD_STATIC=ON -DVCPKG_TARGET_TRIPLET="x64-windows-static-md" -DCMAKE_TOOLCHAIN_FILE="c:\vcpkg\scripts\buildsystems\vcpkg.cmake"  ..
-        cd ${{ github.workspace }}\build && cmake --build . --config Release
+        cd ${{ github.workspace }}\build && cmake --build . --config RelWithDebInfo
 
     - name: Build openmaptiles-compatible mbtiles files of Liechtenstein
       run: |
         Invoke-WebRequest -Uri http://download.geofabrik.de/europe/${{ env.AREA }}-latest.osm.pbf -OutFile ${{ env.AREA }}.osm.pbf
-        ${{ github.workspace }}\build\Release\tilemaker.exe ${{ env.AREA }}.osm.pbf --config=resources/config-openmaptiles.json --process=resources/process-openmaptiles.lua --output=${{ env.AREA }}.mbtiles --verbose || true
+        ${{ github.workspace }}\build\RelWithDebInfo\tilemaker.exe ${{ env.AREA }}.osm.pbf --config=resources/config-openmaptiles.json --process=resources/process-openmaptiles.lua --output=${{ env.AREA }}.mbtiles --verbose || true
 
     - name: 'Upload compiled executable'
       uses: actions/upload-artifact@v2
@@ -45,7 +45,8 @@ jobs:
         name: tilemaker-windows
         path: |
           ${{ github.workspace }}\resources
-          ${{ github.workspace }}\build\Release\tilemaker.exe
+          ${{ github.workspace }}\build\RelWithDebInfo\tilemaker.exe
+          ${{ github.workspace }}\build\RelWithDebInfo\*.pdb
 
   unix-build:  
     strategy:  
--- a/include/attribute_store.h
+++ b/include/attribute_store.h
@@ -92,7 +92,11 @@ struct AttributeStore
 	using key_value_index_t = std::pair<Index, char>; 
 	using key_value_map_t = std::vector< std::pair<std::mutex, key_value_set_t> >;
 
-	key_value_map_t set_list = key_value_map_t(4096);
+	key_value_map_t set_list;
+
+	AttributeStore(unsigned int threadNum) 
+		: set_list(threadNum * threadNum)
+	{ }
 
 	key_value_set_ref_t empty_set() const { return std::make_shared<key_value_set>(); }
 
--- a/include/attribute_store.h
+++ b/include/attribute_store.h
@@ -3,9 +3,7 @@
 #define _ATTRIBUTE_STORE_H
 
 #include "vector_tile.pb.h"
-#include <unordered_set>
-#include <boost/container/small_vector.hpp>
-#include <algorithm>
+#include <boost/functional/hash.hpp>
 
 /*	AttributeStore 
  *	global dictionaries for attributes
@@ -24,6 +22,10 @@ struct AttributeStore
 		std::string key;
 		vector_tile::Tile_Value value;
 		char minzoom;
+
+		kv_with_minzoom(std::string const &key, vector_tile::Tile_Value const &value, char minzoom)
+			: key(key), value(value), minzoom(minzoom)
+		{ }  
 	};
 
 	enum class Index { BOOL, FLOAT, STRING };
@@ -58,7 +60,6 @@ struct AttributeStore
 
 		throw std::runtime_error("Invalid type in attribute store");
 	}
-
     
     struct key_value_less {
         bool operator()(kv_with_minzoom const &lhs, kv_with_minzoom const& rhs) const {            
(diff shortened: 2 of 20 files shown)


systemed:
Cool!

So far it seems a little slower (39m47 on the GB extract vs 34m11) but as it's a WIP I won't take that as a final answer. :)

kleunen:
You use the osm store or not? Maybe try without osm store

kleunen:
I am suspecting that actually this lua interpreter is not multithreaded and locks when executing the script. So different scripts can not be executed in parallel.

systemed:
That's a very good point - http://lua-users.org/wiki/ThreadsTutorial would seem to corroborate that.

I _think_ we could fix that by having a Lua state per thread, rather than one globally:

> Each thread in C which interacts with Lua will need its own Lua state. Each of these states has its own runtime stack. When a new C thread is started, you can create its Lua state in one of two ways.

kleunen:
One way is to call lua_open. This creates a new state which is independent of the states in other threads. In this case, you'll need to initialize the Lua state (for example, loading libraries) as if it was the state of a new program. This approach eliminates the need for mutex locks (discussed below), but will keep the threads from sharing global data.

kleunen:
Have another try, 

also check the number of generated points/lines/polygons. 
I think maybe some lines are missing now ? Not completely sure

systemed:
I’m away for a few days (probably not too feasible to run tilemaker on my phone ;) ) but I’ll have a look when I’m back home.

kleunen:
No problem. The mmap allocator is still preventing it from really scaling up. So. No big improvements yet.

i just wondered why i see a much lower number of linestrings stored. I store the generated geometries now based on osmID. Is it possible the same linestring is generated to different output objects with different attributes? Maybe at different zoom levels?

kleunen:
It works now to some extend. The filesystem backed --store does not work, it crashed. Only the shared memory. But it seems there is some degree of parallel loading now.

kleunen:
Small error in this store code, it is fixed now

kleunen:
On my 12 core machine it nows reads the great britain extract in 3 and and half minute or so and generate the geometries. The ways are most benefiting from parallel reading.

kleunen:
IO is really becoming a bottleneck now. When I do this conversion on my laptop, 12 cores, I see the processing sometimes writing at 1000M/sec - 1200M/sec. It is writing data to store like crazy.

kleunen:
I made some small fixes, but I do notice there are still more polygons generated than before these commits: 

Now (bretagne):
Stored 29831824 nodes, 3956027 ways, 6696 relations
Shape points: 0, lines: 0, polygons: 42
Generated points: 632362, lines: 1232032, polygons: **3189473**

Before:
Stored 29831824 nodes, 3956027 ways, 6696 relations
Shape points: 0, lines: 0, polygons: 42
Generated points: 632362, lines: 1232032, polygons: **3189464**

kleunen:
It is really fast now, even on my small 4 core mobile cpu, reading 4 threads alomst 100% at the same time. The number of generated geometries (lines/polygons) is different than before this commit. I have no idea where this difference is coming from. The result is always the same, independent of how many threads are used. So i guess it is not a threading issue.

systemed:
````
real	24m49.112s
user	323m10.144s
sys	2m28.737s
````

👍 

That's a huge improvement. I'll look further tomorrow (only just got back) but that's a good 10 minutes faster than previously.

kleunen:
Compare also the number of generated geometries. Also look at what the cpu usage is at loading. On my 12 core laptop it is about 90%

kleunen:
Ah ok, i do see a crash on windows executable when reading the relations of my netherlands extract. I have no idea where this is coming from. 

I am running now using a sanitizer, and i did found two other issues. But not the crash yet. It is quite slow now, so I am just leaving it running.

kleunen:
I think i found the crash issue also

systemed:
It looks like all 16 CPUs in use. Memory usage 17.6GB max. Results:

````
Stored 139385816 nodes, 19111319 ways, 65265 relations 
Shape points: 0, lines: 0, polygons: 678
Generated points: 3846401, lines: 11445263, polygons: 10712264
Zoom level 7, writing tile 1060956 of 1060956                
Filled the tileset with good things at ../tm_debug/gb_parallel.mbtiles

real	24m39.443s
user	314m3.797s
sys	2m19.267s

-rw-r--r-- 1 richard richard 1266479104 Jun  2 21:14 ../tm_debug/gb_parallel.mbtiles
````

Compared to master (15.2GB max):

````
Stored 139385817 nodes, 19111319 ways, 65265 relations
Shape points: 0, lines: 0, polygons: 678
Generated points: 3846401, lines: 11445266, polygons: 10712232
Zoom level 14, writing tile 1060956 of 1060956               
Filled the tileset with good things at ../tm_debug/gb.mbtiles

real	34m20.678s
user	311m0.885s
sys	1m38.623s

-rw-r--r-- 1 richard richard 1328132096 Jun  2 23:27 ../tm_debug/gb.mbtiles
````

So it's much faster; there's an increase in RAM used (probably because I was running with `--compact` on master); there's a very slight difference in the number of polygons (which I don't think is significant), and the generated file is a little smaller. I'll eyeball the resulting tiles to see if I can spot any differences.

kleunen:
In your case there is one node less stored. That is a just silly. Where is this one single node. :)

kleunen:
There should also be no reason why the output is smaller. So have a good look at the map. See if there are any differences.

kleunen:
When I look at my netherlands extract: 
Stored 191051541 nodes, 28247961 ways, 106013 relations
Shape points: 0, lines: 0, polygons: 168
Generated points: 15160547, lines: 8783910, polygons: 22598060

And osmium fileinfo:
  Number of nodes: 191051541
  Number of ways: 28247961
  Number of relations: 302886

Some of the relations are not stored.

systemed:
Indeed - we wouldn't expect tilemaker to store all the relations because we don't parse them all (routes, for example). We only store all the ways because they might be used by the relations (and all the nodes because they might be used by the ways).

I ran Europe with this PR overnight. 54GB max memory usage (using store on SSD, obviously) and pretty quick:

IMG_3021

We seem to be closing linestrings when we shouldn't. I think this might be related to simplifying - it only seems to affect geometries that are simplified at the given zoom level.

Screenshot 2021-06-03 at 09 27 47

Screenshot 2021-06-03 at 09 27 07

kleunen:
It might be simplifier, but it might also be the combining. Similar issue was this issue:
https://github.com/systemed/tilemaker/issues/240

kleunen:
Oh wait, i think i know what the issue might be

systemed:
That fixed it!

kleunen:
There is still at least one thing i want to change. So dont merge yet.

kleunen:
Made some more small changes to clean up and improve some small things.

systemed:
Europe with the latest version of this PR:

IMG_3027

Looks good to me! Ready to merge?

toshelp:
It is a rendering result of Sweden, but it seems that some "wood" is lost.
Have any plans to fix this?

coordinate ```7.8/60.123/16.572``` 
sd

kleunen:
> It is a rendering result of Sweden, but it seems that some "wood" is lost.
> Have any plans to fix this?
> 
> coordinate `7.8/60.123/16.572`
> sd

Is the result of this PR or PR 239 ? 
````
diff --git a/include/osm_lua_processing.h b/include/osm_lua_processing.h
index 5911f66..8c1cfa9 100644
--- a/include/osm_lua_processing.h
+++ b/include/osm_lua_processing.h
@@ -125,7 +125,7 @@ public:
         } else if (isWay && !geom::is_valid(geom,failure)) {
             if (verbose) std::cout << "Way " << originalOsmID << " has " << boost_validity_error(failure) << std::endl;
         }
-        if (failure==boost::geometry::failure_self_intersections || failure == boost::geometry::failure_few_points) 
+        if (failure == boost::geometry::failure_few_points) 
                        return false;
                if (failure==boost::geometry::failure_spikes)
                        geom::remove_spikes(geom);

toshelp:
This is the result of this PR.
Wood is not lost in the current master branch.

kleunen:
Hmm, yes, there is something changed which indeed causes certain polygons to go missing. It would be be good to resolve this.

kleunen:
I do see that this init:
https://github.com/systemed/tilemaker/blob/master/src/osm_lua_processing.cpp#L500

Is not performed in setRelation:
https://github.com/systemed/tilemaker/blob/master/src/osm_lua_processing.cpp#L580-L585

systemed:
Not sure - do you have an OSM object ID and a (preferably small :) ) extract where this is consistently happening?

kleunen:
@toshelp Has this island on the right which has a large area of forest is missing. Let me make an extract of this.

kleunen:
[Sweden extract](https://mqtt.kleunen.nl/map/sweden-extract.osm.pbf)

image

kleunen:
https://www.openstreetmap.org/way/34905544#map=12/60.3996/18.4346

kleunen:
It seems the geometry does get created, polygon with 227 nodes, and it is processing in the tile generation phase

systemed:
Found it!

When we call `SetMinZoomByArea` in the Lua script, and that in turn calls `way:Area()`, it's receiving a negative area - -25135521.103994 to be precise.

This causes `SetMinZoomByArea` to write it out with a `minZoom` of 14, so the polygon isn't visible at any lower zoom level.

The [Boost.Geometry docs](https://www.boost.org/doc/libs/1_61_0/libs/geometry/doc/html/geometry/reference/algorithms/area/area_1.html) say that a negative area indicates "Reversed polygon (coordinates not according their orientiation)".

So this suggests that we need to correct the orientation of the polygon before the Lua script gets hold of it (or on the first call to `:Area()`).

kleunen:
Aha. I removed a correct call from the osm store. But the correctgeometry should correct the orientation also right? Or note?

kleunen:
Ah wait. This is not before calling the lua script

systemed:
`:Area()` uses polygonCached, whereas `:Layer()` takes linestringCached, assigns the points to mp, and then corrects _that_. So the original polygonCached is never corrected, even though we'd typically be calling `:Area()` after `:Layer()`.

kleunen:
But then I guess it is best to immediatly call CorrectGeometry in the Cached functions ? 
That way it is always correct when it is used

systemed:
That would make sense!

kleunen:
If i do that it looks completely distorted: 
image

Not sure what is happening there

kleunen:
I will just restore the correct like it was on the current master:
https://github.com/systemed/tilemaker/blob/master/include/osm_store.h#L584

kleunen:
This works fine, so maybe some other time have a good look at this geometry correction :)

systemed:
Excellent.

@toshelp, does this fix things for you?

toshelp:
That's fixed!
Thank you.

kleunen:
cheers

systemed:
Yay!

Thanks again - massive improvement in performance.

I'll update the docs at some point - I slightly miss `--compact` mode but there's probably not much practical difference if people are using an on-SSD store.

kleunen:
It wouldn't be too difficult to readd the compact mode. But it indeed only provides benefit if using the in-memory converter. But i went through the OSMStore and decided to keep the code base as simple as possible. All these option accumulating also makes it difficult to maintain.

systemed:
Yep, I can understand that. Let's leave it out for now and if anyone has an issue I'm sure they'll say so!

PR #245, opened 2021-06-04
Fix some things in PR 239

kleunen:
Simplify fix out-of bounds access
Virtual destructor missing
Don't drop polygon if self-intersecting

### DIFF
--- a/include/osm_lua_processing.h
+++ b/include/osm_lua_processing.h
@@ -125,7 +125,7 @@ public:
         } else if (isWay && !geom::is_valid(geom,failure)) {
             if (verbose) std::cout << "Way " << originalOsmID << " has " << boost_validity_error(failure) << std::endl;
         }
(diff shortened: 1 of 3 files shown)


systemed:
I guess the simplify and self-intersecting changes are already fixed but we still need the virtual destructor fix?

kleunen:
No, because the whole virtual and PbfReaderOutput is removed now in the master. It is using OsmLuaProcessing directly.

systemed:
Ah, hmm. I'm still getting very intermittent segfaults, not sure why.

kleunen:
Hmm, i hope not a threading issue. Maybe run with thread sanitizer:
https://github.com/google/sanitizers/wiki/ThreadSanitizerCppManual

systemed:
It seems to be close to the end of the entire process rather than on .pbf read. I'll keep an eye out and see if it happens again. Will close this for now.

PR #247, opened 2021-06-07
Fix simplification

systemed:
After #239 we're not simplifying as much as we expect (often not at all). I think this is because of a wrong comparison in [simplify](https://github.com/systemed/tilemaker/blob/master/src/geom.cpp#L60). This PR fixes it.

I've also reverted to Boost.Geometry for linestring simplification, as it's valid for linestrings to have self-intersections, and the #239 code with this PR was sometimes causing linestrings to be dropped entirely.

### DIFF
--- a/src/geom.cpp
+++ b/src/geom.cpp
@@ -15,7 +15,7 @@ struct simplify_rtree_counter
 };
 
 template<typename GeometryType>
-void simplify(GeometryType const &input, GeometryType &output, double max_distance, simplify_rtree const &outer_rtree = simplify_rtree())
+void simplify(GeometryType const &input, GeometryType &output, double max_distance, double close_tolerance, simplify_rtree const &outer_rtree = simplify_rtree())
 {        
 	simplify_rtree rtree;
 
@@ -53,11 +53,14 @@ void simplify(GeometryType const &input, GeometryType &output, double max_distan
 			boost::geometry::index::query(outer_rtree, boost::geometry::index::nearest(line, nearest_query_size), std::back_inserter(nearest));
 
 			double min_distance = std::numeric_limits<double>::max();
-			for(auto const &i: nearest) 
-				min_distance = std::min(min_distance, boost::geometry::distance(line, i));
+			for(auto const &i: nearest) {
+				double dist = boost::geometry::distance(line, i);
+				if (dist>0.0) min_distance = std::min(min_distance, dist);
+			}
+		}
 
             std::size_t query_expected = ((start == 0 || end == input.size() - 1) ? 2 : 4);
-            if(result.size() == query_expected && min_distance < max_distance) {
+            if(result.size() == query_expected && min_distance > close_tolerance) {
                 nodes.erase(nodes.begin() + entry + 1);
                 rtree.remove(simplify_segment(input[start], input[middle]));
                 rtree.remove(simplify_segment(input[middle], input[end]));
@@ -74,7 +77,7 @@ void simplify(GeometryType const &input, GeometryType &output, double max_distan
         boost::geometry::append(output, input[i]);
 }
 
-Polygon simplify(Polygon const &p, double max_distance) 
+Polygon simplify(Polygon const &p, double max_distance, unsigned zoom) 
 {
 	simplify_rtree outer_rtree;
 	for(std::size_t j = 0; j < p.outer().size() - 1; ++j) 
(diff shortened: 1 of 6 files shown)


src/geom.cpp
```diff
@@ -57,7 +57,7 @@ void simplify(GeometryType const &input, GeometryType &output, double max_distan
 				min_distance = std::min(min_distance, boost::geometry::distance(line, i));
 
             std::size_t query_expected = ((start == 0 || end == input.size() - 1) ? 2 : 4);
-            if(result.size() == query_expected && min_distance > max_distance) {
+            if(result.size() == query_expected && min_distance < max_distance) {
```
kleunen, line 60: This check was added to ensure that the new line does not get too close to other (non-related) lines of the polygon. Due to rounding of the coordinates, this may result in a self intersection. The check was correct, but maybe max_distance is too strict.

kleunen: For example on these fjords and lochs, you have two sides which are very close to each other. After simplification, the lines of these polygons may be even closer. Not intersecting, but maybe after rounding of the coordinates, they do intersect.

systemed: Ah. If I add this:

> if (min_distance!=0) { std::cout << "min_distance is " << min_distance << std::endl; }

then there's no output... i.e. min_distance is always 0.

kleunen: Yes, i think i know why this is. A query is performed to obtain the 5 nearest lines of the polygon:

kleunen: But that would suggest that indeed the simplifier is not simplifying at all.

systemed: Yes - I think that's the case. Current master at z10 looks like the original geometry:

Screenshot 2021-06-07 at 12 57 56

but with the check removed (i.e. equivalent to this PR):

Screenshot 2021-06-07 at 13 00 21

which is closer to what I'd expect.

kleunen: You can try removing the check, but this may cause issues again on these islands of scotland. That is something to check.

kleunen: I think you can actually remove this whole check, so these lines: 

https://github.com/systemed/tilemaker/blob/fix_simplify/src/geom.cpp#L50-L57

And change:
https://github.com/kleunen/tilemaker/blob/master/src/geom.cpp#L60

from: 
````C++
-if(result.size() == query_expected && min_distance > max_distance) {
````

to:
````C++
+if(result.size() == query_expected) {
````

I checked on this scotland export, it seems to work fine even without this distance check. North Uist is not displayed at zoom level 4, but is displayed at other zoom levels. I think this is how we eventually accepted the PR.

**Comment by systemed** [comment]
I've pushed a further change - any thoughts? It retains the rtree check, but rather than using max_distance, has a (zoom-dependent) value to make sure lines don't get too close together.

North Uist is still disappearing at z4 but otherwise it seems to work fine.

File: src/geom.cpp [hunk 2]
```diff
@@ -53,11 +53,14 @@ void simplify(GeometryType const &input, GeometryType &output, double max_distan
...
-				min_distance = std::min(min_distance, boost::geometry::distance(line, i));
+			for(auto const &i: nearest) {
+				double dist = boost::geometry::distance(line, i);
+				if (dist>0.0) min_distance = std::min(min_distance, dist);
+			}
+		}
 
             std::size_t query_expected = ((start == 0 || end == input.size() - 1) ? 2 : 4);
-            if(result.size() == query_expected && min_distance < max_distance) {
+            if(result.size() == query_expected && min_distance > close_tolerance) {
```
kleunen, line 63: Nearest tolerance instead of close tolerance?

systemed: Yep, that would work as a name!

kleunen:
> I've pushed a further change - any thoughts? It retains the rtree check, but rather than using max_distance, has a (zoom-dependent) value to make sure lines don't get too close together.
> 
> North Uist is still disappearing at z4 but otherwise it seems to work fine.

i think you can just remove this check completely. If the two lines are within the tolerance of the zoom level, after rounding they will touch. But not really intersect. Renders like mapbox can handle this i think. And you are not going to see the difference on the map if the lines are this close to each other.

systemed:
Yep, that works fine. 👍 

There's apparently a tilemaker talk at the FOSSGIS conference tomorrow so I'll merge this before then! https://www.fossgis-konferenz.de/2021/sessions/Y8RGMT.php

systemed:
Successfully generated Europe overnight with this - no difference in timing (still 492 minutes). mbtiles size is down from 26GB to 24.75GB, most of which is in the lower zooms.

I think there's still a bit of config file optimisation we can do - tile 4/8/5 is massive at 744k, I suspect due to unsimplified borders. But codewise this will do.

systemed:
It's a bug in `ReorderMultiLinestring` - if I comment that out then the output is expected. Haven't found what yet!

systemed:
Should be fixed in #248.

PR #248, opened 2021-06-08
Limit combined linestrings to 6000 vertices

systemed:
This addresses #240 - seems to fix it for me. Comparing file sizes:

- current master: 749568 (rendering error)
- disabling ReorderMultiLinestring: 757760 (rendering ok but linestrings are disconnected, hence larger filesize)
- this PR: 749568 (rendering correct, smaller filesize)

My working assumption is that MBGL can't cope with the very long linestrings created at low zoom levels when the coastline is merged together. This PR limits linestrings to 6000 vertices.

There is no formal limit in the MVT specification, but https://docs.mapbox.com/mapbox-tiling-service/reference/ says "complex features will receive additional simplification so that they can be rendered correctly by Mapbox GL. Polygons or MultiPolygons with more than 65535 vertices and LineStrings and polygon rings with more than 6553 points will be simplified to reduce their complexity to these limits", so it's a known issue. I'm not sure whether 6553 is a typo for 65535, but 6000 is plenty anyway!

### DIFF
--- a/src/tile_worker.cpp
+++ b/src/tile_worker.cpp
@@ -35,7 +35,7 @@ void ReorderMultiLinestring(MultiLinestring &input, MultiLinestring &output) {
 			auto foundStart = startPoints.find(xy_pair(lastPoint.x(),lastPoint.y()));
 			if (foundStart != startPoints.end()) {
 				unsigned idx = foundStart->second;
-				if (!added[idx]) {
+				if (!added[idx] && input[idx].size()+ls.size()<6000) {
 					ls.insert(ls.end(), input[idx].begin()+1, input[idx].end());
 					added[idx] = true;
 					continue;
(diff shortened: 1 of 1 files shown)


kleunen:
@QuentinC Can you verify this works for you on your France extract ?

QuentinC:
> 
> 
> @QuentinC Can you verify this works for you on your France extract ?

I'm currently running a planet generation, I should be able to test this on the end of the week !
I'll keep you informed !

QuentinC:
> 
> 
> @QuentinC Can you verify this works for you on your France extract ?

Hello,

I can confirm that this fixes the issue with the coastline shp file !
Here is the result with the original config (using combine_below): [testshp.zip](https://github.com/systemed/tilemaker/files/6641035/testshp.zip)

Thanks !

Issue #242, opened 2021-05-22
Lost polygon after #239


QuentinC:
Hello,

During my tests on this software, I have seen a lost polygon. 
It displays fine before #239 but disappears after...
This is the lost polygon: https://www.openstreetmap.org/relation/2980911 

Well it might be more complicated, here are some screenshots taken before #239 :
Zoom 12:
![1](https://user-images.githubusercontent.com/3394508/119241260-666d8580-bb55-11eb-9b6e-c2451d39ddbe.jpg)
Zoom 13:
![2](https://user-images.githubusercontent.com/3394508/119241265-6cfbfd00-bb55-11eb-8e65-5e91fc44c8b4.jpg)
And Zoom 14:
![14](https://user-images.githubusercontent.com/3394508/119241269-738a7480-bb55-11eb-939f-057b78d65796.jpg)

As you can see some parts of the polygon are lost depending of the zoom level, and the whole polygon disappear when using latest commit... 
I have dont my tests using the France>Bretagne export from there: https://download.geofabrik.de/europe/france/bretagne-latest.osm.pbf with a custom process.lua that includes military landuse:
[process-openmaptiles.lua.txt](https://github.com/systemed/tilemaker/files/6527194/process-openmaptiles.lua.txt) (L496)

Thanks again for the support !

toshelp:
```is_valid``` checks geometry in the sense of OCG. I think this is the result of that check.
https://www.boost.org/doc/libs/1_76_0/libs/geometry/doc/html/geometry/reference/algorithms/is_valid/is_valid_2_with_failure_value.html

I found a workaround to uncheck ```failure_self_intersections```. But, it may cause other problems.
```
diff --git a/include/osm_lua_processing.h b/include/osm_lua_processing.h
index 5911f66..8c1cfa9 100644
--- a/include/osm_lua_processing.h
+++ b/include/osm_lua_processing.h
@@ -125,7 +125,7 @@ public:
         } else if (isWay && !geom::is_valid(geom,failure)) {
             if (verbose) std::cout << "Way " << originalOsmID << " has " << boost_validity_error(failure) << std::endl;
         }
-        if (failure==boost::geometry::failure_self_intersections || failure == boost::geometry::failure_few_points) 
+        if (failure == boost::geometry::failure_few_points) 
                        return false;
                if (failure==boost::geometry::failure_spikes)
                        geom::remove_spikes(geom);

```

kleunen:
[...]

PR #249, opened 2021-06-09
Perform dissolve on self-intersecting polygons

kleunen:
Have a try to see how it works, on my 'Deventer' export, it actually helps. This polygon which was causing rendering issues on the tile, it is repaired and no rending issue.

### DIFF
--- a/include/dissolve/dissolve.hpp
+++ b/include/dissolve/dissolve.hpp
@@ -1,0 +1,235 @@
+#ifndef __BOOST_GEOMETRY_DISSOLVE_H__
+#define __BOOST_GEOMETRY_DISSOLVE_H__
+
+/*
+ * ----------------------------------------------------------------------------
+ * "THE BEER-WARE LICENSE" (Revision 42):
+ * Wouter van Kleunen wrote this file.  As long as you retain this notice you
+ * can do whatever you want with this stuff. If we meet some day, and you think
+ * this stuff is worth it, you can buy me a beer in return. 
+ * ----------------------------------------------------------------------------
+ */
+
+#include <vector>
+#include <boost/geometry.hpp>
+#include <boost/geometry/geometries/point_xy.hpp>
+#include <boost/geometry/geometries/polygon.hpp>
+#include <boost/geometry/geometries/multi_polygon.hpp>
+
+namespace dissolve {
+
+template<typename C, typename T>
+static inline void result_combine(C &result, T &&new_element)
+{
+    result.push_back(new_element);
+
+   	for(std::size_t i = 0; i < result.size() - 1; ) {
+        if(!boost::geometry::intersects(result[i], result.back())) {
+            ++i;
+            continue;
+        }
+
+        std::vector<T> union_result;
+        boost::geometry::union_(result[i], result.back(), union_result);
+
+        if(union_result.size() != 1) {
+			++i;
+			continue;
+		}
+
+       	result.back() = std::move(union_result[0]);
+       	result.erase(result.begin() + i);
+    } 
+}
+
+struct pseudo_vertice_key
+{
+    std::size_t index_1;
+    double scale;
+    std::size_t index_2;
+    bool reroute;
+    
+    pseudo_vertice_key(std::size_t index_1 = 0, std::size_t index_2 = 0, double scale = 0.0, bool reroute = false)
+        : index_1(index_1), scale(scale), index_2(index_2), reroute(reroute)
+    { } 
+};
+
+struct compare_pseudo_vertice_key
+{
+    bool operator()(pseudo_vertice_key const &a, pseudo_vertice_key const &b) const {
+        if(a.index_1 < b.index_1) return true;
+        if(a.index_1 > b.index_1) return false;
+        if(a.scale < b.scale) return true;
+        if(a.scale > b.scale) return false;
+        if(a.index_2 > b.index_2) return true;
+        if(a.index_2 < b.index_2) return false;
+        if(a.reroute && !b.reroute) return true;
+        if(!a.reroute && b.reroute) return false;
+		return false;
+    }
+};
+
+template<typename point_t = boost::geometry::model::d2::point_xy<double>>
+struct pseudo_vertice
+{
+    point_t p;
+    pseudo_vertice_key link;
+    
+    pseudo_vertice(point_t p, pseudo_vertice_key link = pseudo_vertice_key())   
+        : p(p), link(link)
+    { }        
+};
+
+template<
+	typename point_t = boost::geometry::model::d2::point_xy<double>, 
+	typename ring_t = boost::geometry::model::ring<point_t>
+	>
+static inline void dissolve_find_intersections(
+			ring_t const &ring,
+			std::map<pseudo_vertice_key, pseudo_vertice<point_t>, compare_pseudo_vertice_key> &pseudo_vertices,
+    		std::set<pseudo_vertice_key, compare_pseudo_vertice_key> &start_keys)
+{
+	// Generate all by-pass intersections in the graph
+	// Generate a list of all by-pass intersections
+    for(std::size_t i = 0; i < ring.size(); ++i)
+    {
+        pseudo_vertices.emplace(pseudo_vertice_key(i, i, 0.0), ring[i]);       
+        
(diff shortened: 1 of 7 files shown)


systemed:
I tried this last night and it seems to work really well!

If you patch it into https://github.com/systemed/tilemaker/blob/master/src/read_shp.cpp#L248 then it can fix geometries from shapefiles too - there's a faulty one in the coast around Britain which this successfully fixes.

kleunen:
I will do that hopefully tonight.

kleunen:
Have a try

systemed:
Working well on the UK and France too - I tried it on the France extract (which has eight gazillion buildings) and it worked well. In particular, the lost polygon in Brest from #242 seems to be fixed!

Only issue I had was that it terminated badly with france-latest.osm.pbf:

````
richard@peyresourde:~/tilemaker$ time tilemaker --input /media/data1/planet/france-latest.osm.pbf \
--output /media/data3/france.mbtiles \
--process resources/process-openmaptiles.lua --config resources/config-openmaptiles.json \
--store /media/ssd/

[...]

Shape points: 0, lines: 0, polygons: 941
Generated points: 9311283, lines: 17823864, polygons: 51471225
Zoom level 8, writing tile 574799 of 574799                
Filled the tileset with good things at /media/data3/france.mbtiles
terminate called without an active exception
Aborted (core dumped)

real	33m52.533s
user	319m30.878s
sys	7m1.917s
````

This was after the .mbtiles had been written so no real effect, but still I wonder why it's doing that.

**Comment by kleunen** [comment]
Probably in the deallocation of memory with this mmap allocator. At exit all shared pointers and data structures are released

**Comment by kleunen** [comment]
Maybe it attempts to delete the directory before all deallocation is done. You can try disabling deallocation in the memory manager.

**Comment by kleunen** [comment]
I think the polygon from #242 is not visible with the default style. It is landuse military. This is not exported with openmaptiles process?

**Comment by kleunen** [comment]
I also pushed a fix for closing the mmap file in this PR, and now perform all corrections using the dissolve.

**Comment by QuentinC** [comment]
> 
> 
> I think the polygon from #242 is not visible with the default style. It is landuse military. This is not exported with openmaptiles process?

Hello,

You are right, here is my adapted process file that includes military:
```
        -- Set 'landuse'
        else
                if l=="" then l=amenity end
                if l=="" then l=tourism end
                if landuseKeys[l] then
                        way:Layer("landuse", true)
                        way:Attribute("class", l)
                        if l=="residential" then
                                if way:Area()<ZRES8^2 then way:MinZoom(8)
                                else SetMinZoomByArea(way) end
                        else way:MinZoom(11) end
                        write_name = true
                end
                if l=="military" then
                        SetMinZoomByArea(way)
                        write_name = false

                        way:LayerAsCentroid("landuse_detail")
                        SetNameAttributes(way)
                        way:Attribute("class", l)
                        SetMinZoomByArea(way)
                end
        end
```

I am on my way to run this PR to check if it helps !

QuentinC:
I will have to do some more investigation, right now it seems that my way is still having problems.

I've seen something strange however, my run time went from 8 to 47min ! It took a lot of time reading a shapefile (ne_10m_ocean.shp).
It might be related to https://github.com/systemed/tilemaker/pull/249/commits/7f2a12d5b5c1fbae4ab8ce43e72eb1c1aa1a4925 ?

I will do some more tests on this on the next days and I will comment here !

kleunen:
Yes, the dissolve can take a lot of time. Also, boost geometry correct takes a lot of time, i think most of the time of the input process. I profiled this once. But that is a big difference you mention.

kleunen:
Boost geometry is actually used by mysql:
https://mysqlserverteam.com/why-boost-geometry-in-mysql/

I found a dataset invalid polygons: 
https://stackoverflow.com/questions/49902090/dataset-of-invalid-geometries-in-boostgeometry

And added this to my boost/geometry/correct:
https://github.com/kleunen/boost_geometry_correct

Would be nice if it someday would become a mysql makevalid

kleunen:
I am using an rtree for calculating the intersections now, this should be faster

systemed:
Latest code run successfully with both UK and France:

````
real	24m39.318s
user	306m45.536s
sys	2m36.786s

real	33m44.434s
user	317m26.253s
sys	6m56.542s
````

With France I was using the mmap store and got an exception (Boost 1.71) at the end of processing:

````
terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
  what():  boost::filesystem::remove: Permission denied: "/media/ssd/"
Aborted (core dumped)
````

This doesn't affect anything because the tiles were already generated; there seems to be a bit of commentary on StackOverflow etc. from other people who've encountered this, so maybe just catching that error would work.

It might be good to apply the dissolve fix to `calculateCentroid` in osm_lua_processing.cpp too - failing to get centroids of invalid polygons is quite a common error, e.g.:

    Problem geometry relation 10316772: Boost.Geometry Centroid calculation exception

But that's not urgent and we could do it in a later PR.

kleunen:
> With France I was using the mmap store and got an exception (Boost 1.71) at the end of processing:
> 
> ```
> terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
>   what():  boost::filesystem::remove: Permission denied: "/media/ssd/"
> Aborted (core dumped)
> ```

What was the path that you used for the osm_store ?

systemed:
Just `/media/ssd/`.

kleunen:
ah ok, yes, remove_all tries to remove all files/directories below /media/ssd and /media/ssd itself. Maybe it would be good not to use remove_all, but rather track the files that are actually created. I am bit scared tilemaker is going to delete any files of the user on disk...

QuentinC:
That might be because /media/ssd is a mount point ?

systemed:
I think this is good to merge - shall we go for it?

kleunen:
There is interest from boost geometry to add a feature like this to boost, have a look at the discussion here:
https://github.com/boostorg/geometry/issues/868#issuecomment-864564249

I have made some extra improvements to the correct, based on this. I can merge them into this PR. I don't think you will need all improvements that will come out of the boost/geometry discussion. They are probably very specific details. 

There is still one change I would like to merge, so please wait with merging :)

kleunen:
Have a try, i pushed this last small change :)

kleunen:
Moving towards 2.0 release, i would say ?

systemed:
Yes, definitely. I've set a deadline of State of the Map 2021 (9-11 July) for 2.0 :)

Just running a final test and will merge this afterwards, all things being equal!

kleunen:
can you still host a workshop there ? or is too late to enter ?

systemed:
Too late for a workshop, but I've signed up for a lightning talk: https://wiki.openstreetmap.org/wiki/State_of_the_Map_2021/registration_lightning_talks

systemed:
We're good to go! Merged. Thank you for an absolutely outstanding contribution.

Really good to see the interest from Boost.Geometry too!

kleunen:
Yes, it will probably be a long process with boost geometry. 
But hopefully someday it will be available for many other projects as well.

kleunen:
I think commit db992ce4f49541200fc440d56bcbba1901357dbd limits the performance quite severly, you might want to revert this

kleunen:
> 
> I've seen something strange however, my run time went from 8 to 47min ! It took a lot of time reading a shapefile (ne_10m_ocean.shp).
> It might be related to [7f2a12d](https://github.com/systemed/tilemaker/commit/7f2a12d5b5c1fbae4ab8ce43e72eb1c1aa1a4925) ?
> 
> I will do some more tests on this on the next days and I will comment here !

@QuentinC can you try if the performance is still acceptable ? If the performance is not back to very slow again ?

QuentinC:
> 
> 
> > I've seen something strange however, my run time went from 8 to 47min ! It took a lot of time reading a shapefile (ne_10m_ocean.shp).
> > It might be related to [7f2a12d](https://github.com/systemed/tilemaker/commit/7f2a12d5b5c1fbae4ab8ce43e72eb1c1aa1a4925) ?
> > I will do some more tests on this on the next days and I will comment here !
> 
> @QuentinC can you try if the performance is still acceptable ? If the performance is not back to very slow again ?

It seems to be running fine here, last try has been done in 9min so no issue for me !
I will have a look at https://github.com/systemed/tilemaker/pull/258 !

PR #255, opened 2021-06-18
Keep parts of docker build and small optimisation in tile rendering

QuentinC:
Hello,

Here is a very small contribution in this project:
- First part is in the Dockerfile, when making tests it's faster to keep APT dependencies between builds
- When rendering tiles, lower the tiles per thread on low zoom so that 1 CPU is not doing all the massive tiles

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,14 +3,15 @@ LABEL Description="Tilemaker" Version="1.4.0"
 
 ARG DEBIAN_FRONTEND=noninteractive
 
-COPY . /
-WORKDIR /
-
 # install dependencies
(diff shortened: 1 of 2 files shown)


kleunen:
Looks useful :)

systemed:
I'd call that better than a "small optimisation"...  Before (France extract):

````
real	34m40.225s
user	320m20.525s
sys	7m10.526s
````

After:

````
real	25m15.553s
user	342m58.381s
sys	6m50.409s
````

I'd noticed that it sits on one thread at the end of generating the France extract - this is a smart way to tackle that. Thanks!

PR #256, opened 2021-06-19
OSM Store remove only self-generated files and catch errors

kleunen:
Only remove files if they are actually generated by tilemaker

### DIFF
--- a/src/osm_store.cpp
+++ b/src/osm_store.cpp
@@ -21,6 +21,7 @@ struct mmap_file
 
 	static size_t mmap_file_size;
 	static std::string mmap_dir_filename;
+	static bool mmap_dir_created;
 	std::string filename;
 
 	boost::interprocess::file_mapping mapping;
@@ -64,6 +65,7 @@ thread_local mmap_file_ptr mmap_file_thread_ptr;
 
 size_t mmap_file::mmap_file_size = 0;
 std::string mmap_file::mmap_dir_filename;
+bool mmap_file::mmap_dir_created = false;
      
 std::vector< mmap_shm_ptr > mmap_shm_regions;  
 size_t mmap_shm::mmap_file_size = 0;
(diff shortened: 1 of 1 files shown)


systemed:
I've resolved (hopefully) the conflicts but there's a couple of compilation warnings:

````
src/osm_store.cpp:28:47: warning: ‘mmap_file::buffer’ will be initialized after [-Wreorder]
   28 |  boost::interprocess::managed_external_buffer buffer;
      |                                               ^~~~~~
src/osm_store.cpp:24:14: warning:   ‘std::string mmap_file::filename’ [-Wreorder]
   24 |  std::string filename;
      |              ^~~~~~~~
src/osm_store.cpp:74:1: warning:   when initialized here [-Wreorder]
   74 | mmap_file::mmap_file(std::string const &filename, std::size_t offset)
      | ^~~~~~~~~
````

It's not removing everything in the directory (which is good!) but it's still coming up with a "Permission denied" error:

````
richard@peyresourde:~/tilemaker$ touch /media/ssd/tmp.txt
richard@peyresourde:~/tilemaker$ tilemaker --input /media/data1/planet/oxfordshire-latest.osm.pbf --output ../tm_debug/oxfordshire.mbtiles --store /media/ssd --process resources/process-openmaptiles.lua --config resources/config-openmaptiles.json 
[...]
Filled the tileset with good things at ../tm_debug/oxfordshire.mbtiles
boost::filesystem::remove: Permission denied: "/media/ssd"
richard@peyresourde:~/tilemaker$ ls /media/ssd/
tmp.txt
````

kleunen:
Yes, the error is still printed. I guess only way to prevent is to detect if directory already exists on startup and not remove on shutdown if it was not created.

kleunen:
Not sure what I pushed now and if this still merges

systemed:
That seems to work!

kleunen:
@systemed  Can this be merged ? I am still a bit concerned tilemaker might delete some users files.

systemed:
Sorry, been a bit behind on things recently! Thank you, happily merged.

PR #258, opened 2021-06-23
Fix performance regression in polygon make_valid

kleunen:
I made a performance regression in that last commit in PR #249 
this should fix this


systemed:
Cool - performance seems similar to me but I hadn't encountered the regression. @QuentinC does this work for you?

kleunen:
I guess it is only noticable when you perform make_valid on very large polygons, with like 10.000 nodes.

kleunen:
If there is no noticable performance difference. Then there is no need to merge this :)

PR #259, opened 2021-07-02
Allow to manually set bbox if .osm.pbf does not contain it

Nakaner:
Planet dumps do not contain a bounding box header and Tilemaker requires one for shapefile import. This pull request adds a command line option to specify the bounding box for this use case.

### DIFF
--- a/src/tilemaker.cpp
+++ b/src/tilemaker.cpp
@@ -15,6 +15,7 @@
 #include <chrono>
 
 // Other utilities
+#include <boost/algorithm/string.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/program_options.hpp>
 #include <boost/variant.hpp>
@@ -119,6 +120,16 @@ void WriteFileMetadata(rapidjson::Document const &jsonConfig, SharedData const &
 	fclose(fp);
 }
 
+double bboxElementFromStr(const string& number) {
+	char* endptr;
+	double num = strtod(number.c_str(), &endptr);
+	if (*endptr != '\0') {
+		cerr << "Failed to parse coordinate " << number << endl;
+		exit(1);
+	}
+	return num;
+}
+
 /**
  *\brief The Main function is responsible for command line processing, loading data and starting worker threads.
  *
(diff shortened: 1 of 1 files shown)


src/tilemaker.cpp
```diff
@@ -163,6 +176,15 @@ int main(int argc, char* argv[]) {
 	if (vm.count("output")==0) { cerr << "You must specify an output file or directory. Run with --help to find out more." << endl; return -1; }
 	if (vm.count("input")==0) { cout << "No source .osm.pbf file supplied" << endl; }
 
+	vector<string> bboxElements;
```
kleunen, line 179: put this inside a function (parsebbox?), which returns std::vector<double>, and uses the lexical_cast. 
Because the main is already quite large

src/tilemaker.cpp
```diff
@@ -119,6 +120,16 @@ void WriteFileMetadata(rapidjson::Document const &jsonConfig, SharedData const &
 	fclose(fp);
 }
 
+double bboxElementFromStr(const string& number) {
+	char* endptr;
+	double num = strtod(number.c_str(), &endptr);
```
kleunen, line 125: Use a lexical_cast for this: 
https://www.geeksforgeeks.org/boost-lexical_cast-in-cpp/

src/tilemaker.cpp
```diff
@@ -193,6 +215,13 @@ int main(int argc, char* argv[]) {
 	Box clippingBox;
 	MBTiles mapsplitFile;
 	double minLon=0.0, maxLon=0.0, minLat=0.0, maxLat=0.0;
```
kleunen, line 217: maybe change this to std::vector<double> ? or struct of type bbox: struct bbox { double minLon; double maxLon; ... }. 
such that parsebbox can return this.

Nakaner: The changes following your suggested change extend the scope of this pull request quite a lot. That's why I decided to skip this for the time being.

systemed:
Thanks!

PR #263, opened 2021-07-05
Add render_height and render_min_height to buildings for 3d rendering

keichan34:
The building levels -> meters conversion uses the value from [building/building.sql](https://github.com/openmaptiles/openmaptiles/blob/master/layers/building/building.sql#L101-L102)


systemed:
Only a 1% filesize increase on a (very building-heavy) France extract, and 0.2% on the (less building-heavy) Oxfordshire extract, so this is great. Thank you!

PR #267, opened 2021-07-10
Close mmap file before deleting

kleunen:
On windows platform the osm_store files were not closed before deletion. 
On linux this works fine, on windows this causes an error message that the file could not be removed. 

I also enabled osm_store on windows test run now, in CI


systemed:
I don't have a Windows machine to test this on, but it builds happily and I'm sure you're right! Thanks.

kleunen:
It seems some changes are now reverted in this PR

systemed:
Aaah. Yes, this appears to have conflicted with #286.

PR #271, opened 2021-07-13
Clamp lat to range valid for latp

kleunen:
Clamp lat to -85.06 to prevent merging artifacts


systemed:
Excellent - thank you.

PR #279, opened 2021-08-05
Improved simplify performance

kleunen:
Improved performance for intersection aware simplifier. 
Also, condition in which intersection may still occur was fixed.

### DIFF
--- a/src/geom.cpp
+++ b/src/geom.cpp
@@ -6,37 +6,30 @@
 
 #include "geometry/correct.hpp"
 
+#include <boost/range/adaptor/transformed.hpp>
+#include <boost/range/irange.hpp>
+
 typedef boost::geometry::model::segment<Point> simplify_segment;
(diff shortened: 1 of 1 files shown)


systemed:
This builds and runs fine for me on macOS, and the 16.04 CI is fine, but for some reason the CI is failing to build on the mbtiles action:

````
  src/geom.cpp:85:5:   required from here
  /usr/include/boost/optional/optional.hpp:733:64: error: use of deleted function 'simplify(const Polygon&, double)::<lambda(std::size_t)>& simplify(const Polygon&, double)::<lambda(std::size_t)>::operator=(const simplify(const Polygon&, double)::<lambda(std::size_t)>&)'
       void assign_value ( rval_reference_type val ) { get_impl() = static_cast<rval_reference_type>(val); }
  src/geom.cpp:83:37: note: a lambda closure type has a deleted copy assignment operator
     | boost::adaptors::transformed([&p](std::size_t i) {
````

kleunen:
I will have a look at this

kleunen:
It compiles now :)

PR #283, opened 2021-08-13
Add z_order sorting support

Nakaner:
This pull request adds z_order support. The Lua script can set a numeric z_order per feature and Tilemaker will sort the features within a layer by z_order (z_order is more important than geometry type).

z_order sorting is needed for map rendering engines which cannot sort features themselves (e.g. Mapnik with OGR plugin). Sorting is limited to features within a layer. This means, if layers are merged with the `write_to` configuration option, sorting does not happen accross the boundaries of the source layers.

This pull request adds z_order support to the OpenMapTiles configuration that comes with Tilemaker. Upstream OpenMapTiles sorts features in transportation layer by z_order (which is calculated by Imposm).


systemed:
This looks terrific - thank you. I’m away from my development machine at the moment but will look/test when I’m back.

systemed:
Merged - thanks!

In the future I might look at the instance variable sizes in OutputObject as I suspect they could be packed more efficiently, and that might have an impact on the range available for z_order. But that will take a bit of investigation so let's run with this for now.

PR #286, opened 2021-08-20
Return compact mode

kleunen:
The following PR returns the compact mode for storing the nodes (reduced memory consumption). 

Use osmium renumber to renumber the nodes:
osmium renumber --input planet-latest.osm.pbf --output planet-renum.osm.pbf

Use compact mode switch to use the compact mode:
./tilemaker --input planet-renum.osm.pbf --output planet.mbtiles --store osm_store --compact

### DIFF
--- a/include/osm_store.h
+++ b/include/osm_store.h
@@ -132,6 +132,63 @@ private:
 	std::shared_ptr<map_t> mLatpLons;
 };
 
+class CompactNodeStore
+{
+
+public:
+	using element_t = std::pair<NodeID, LatpLon>;
+	using map_t = std::deque<LatpLon, mmap_allocator<LatpLon>>;
+
+	void reopen()
+	{
+		std::lock_guard<std::mutex> lock(mutex);
+		mLatpLons = std::make_unique<map_t>();
+	}
+
+	// @brief Lookup a latp/lon pair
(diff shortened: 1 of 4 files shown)


systemed:
Looks good to me - will try it out tomorrow/Sunday!

kleunen:
Non-Compact:
Memory used: 2535272
Filled the tileset with good things at overijssel.mbtiles
real    2m14.339s
user    7m19.937s
sys     0m10.562s

Compact: 
Memory used: 2459180
Filled the tileset with good things at overijssel.mbtiles
real    2m8.946s
user    7m4.379s
sys     0m9.507s

It is only a small extract, so no memory pressure.

systemed:
Seems to work really well! I almost managed to get the planet working with it (144GB RAM) but oomkiller killed it with 99% of the pbf read. :(

kleunen:
Yes. I ran into the same issue when converting africa on my 8G machine. Increase the amount of swap ...

systemed:
Anyway, this is great - thanks. I have a couple more ideas for reducing memory usage I might try.

kleunen:
Maybe make an issue for collecting some ideas of how to reduce memory ?

PR #289, opened 2021-08-23
Changes from #267 which git appears to have missed

systemed:
Who knows what was going on there, but this is `osm_store.cpp` from #267 again.

### DIFF
--- a/src/osm_store.cpp
+++ b/src/osm_store.cpp
@@ -18,7 +18,6 @@ struct mmap_file
 {
 	std::string filename;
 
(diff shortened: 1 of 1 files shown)


kleunen:
It seems the current head is correct, no changes missed ..

kleunen:
Ah yes, mmap_files was removed in the other commit, so there is a conflict.

kleunen:
I will make a PR to fix the compile/merge error

kleunen:
See #290

PR #290, opened 2021-08-23
Fix compile error after merge

kleunen:
Compile fix for changes from #267


systemed:
Works for me - thanks!

Issue #293, opened 2021-08-23
Ideas for further reducing memory usage


systemed:
A few possibilities and thoughts:

### Don't add ways to waystore unless they're needed by relations

See #292.

### Lazy geometry assembly

At present we create our geometries whenever our Lua process asks for an object to be written. This is in `linestringCached`, `polygonCached` and `multiPolygonCached` in osm_lua_processing.cpp.

We could consider creating them lazily, i.e. only when writing each tile. The advantage of this is that we could purge the geometries when memory is getting low; they could subsequently be recreated next time we need them.

One challenge would be calculating the bbox if we haven't generated the geometry yet. A naive approach would simply take the min/max of the node coordinates - not sure whether this would suffice in the case of large geometries.

(This would rule out the approach in #292, because we'd need the waystore to stick around so that we could generate the geometries on demand.)

### Use rtree for OutputObjects

Rather than adding each OutputObject to a list for each z14 tile, we could store an rtree of OutputObjects' bounding boxes, and then query the rtree when writing each tile. (This may need some thought for large, irregularly shaped objects. Performance might be an issue too.)

### Support osmium add-locations-to-ways

This tool pre-processes the .pbf to add node lon/lat to the ways themselves. As a result, you don't need to read them into a node store. I've added support for this to my OSRM fork and it's fairly straightforward. ([Explanation](https://blog.jochentopf.com/2016-04-20-node-locations-on-ways.html), [docs](https://docs.osmcode.org/osmium/latest/osmium-add-locations-to-ways.html).)

### Optimise bitpacking in OutputObject

This isn't too efficient at present (8,8,8,1,4). We could reduce geomType to 2 bits and maybe z_order to 6. Not sure how much of a difference this would make - possibly none if the following variables are aligned to 4-byte boundaries, for example.

### Support a "split layer" or "split region" approach

People have reported success splitting the planet into parts, running tilemaker over each one, and merging the resulting mbtiles. Alternatively, filter the planet by thematic layer, run tilemaker and merge. We could provide tooling to support this - it doesn't have to be in the tilemaker executable, it could be a bash/Ruby/Lua/whatever script that invokes tilemaker, osmium etc. as required.

### Look at shapefile reading

A good starting point might be looking at the result of reading in shapefiles - it looks to me as if we're using many GB to read the coastlines for the whole world, when the shapefile itself is only around 1GB total.

kleunen:
[...]

PR #294, opened 2021-08-24
Pack OutputObject bitfields more efficiently

systemed:
Restricting `objectID` to 40 bits as per #293 means we can fit the rest of the flags into the remaining 24 bits.

Edit: this works fine on clang, but GCC appears to squawk at the bitfield not being big enough for the uint8_t underlying the enum, even though we only have 2 bits worth of enum values.

Edit 2: we may need to look at newWayID/MAX_WAY_ID too.

### DIFF
--- a/include/geom.h
+++ b/include/geom.h
@@ -43,7 +43,7 @@ typedef boost::geometry::index::rtree< IndexValue, boost::geometry::index::quadr
 typedef uint64_t NodeID;
 typedef uint64_t WayID;
 
-#define MAX_WAY_ID std::numeric_limits<WayID>::max()
+#define MAX_WAY_ID pow(2,40)-1
 typedef std::vector<NodeID> NodeVec;
 typedef std::vector<WayID> WayVec;
 
--- a/include/output_object.h
+++ b/include/output_object.h
@@ -39,7 +39,7 @@ protected:
 
 
 public:
-	NodeID objectID 			: 40;					// id of way (linestring/polygon) or node (point)
+	NodeID objectID 			: 40;					// id of way (linestring/polygon) or node (point) - cf MAX_WAY_ID
 	uint_least8_t layer 		: 8;					// what layer is it in?
 	int8_t z_order				: 8;					// z_order: used for sorting features within layers
 	OutputGeometryType geomType : 2;					// point, linestring, polygon
--- a/include/output_object.h
+++ b/include/output_object.h
@@ -29,6 +29,7 @@ std::ostream& operator<<(std::ostream& os, OutputGeometryType geomType);
  * Possible future improvements to save memory:
  * - use a global dictionary for attribute key/values
 */
+#pragma pack(push, 4)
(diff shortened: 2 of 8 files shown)


kleunen:
Change the 'enum class' to an 'enum' and just store in an 'uint8_t : 2' in the outputobject. You lose the type checking, but ok ..

kleunen (commented):
OUTPUTGEOMETRY_POINT, OUTPUTGEOMETRY_LINESTRING, OUTPUTGEOMETRY_POLYGON ?

Or:

OG_POINT, OG_LINESTRING, OG_POLYGON

Or:

GEOMETRY_POINT, GEOMETRY_LINESTRING, GEOMETRY_POLYGON 

:)

systemed:
I did wonder about GEOMTYPE_POINT but I'm a lazy typist ;)

kleunen:
I think at the moment you are not actually saving any bytes on the outputobject. Because the reference count (uint32_t) was combined with the flags before. And now, reference count is still uint32_t, but is stored aligned on 64bit alignment (uint64_t).

systemed:
Gah, yes, you're right. 40 bytes each time.

I'm not sure if it would be possible to replace AttributeStoreRef with an intrusive_ptr and a reference count in the spare four bytes... sounds slightly plausible but it's way beyond my C++-fu :)

kleunen:
Maybe enable packing on the struct. 
pragma pack

I can look it up later

kleunen:
Or replace attributestoreref with an 32 bit attributestoreindex. Not sure if that is possible

kleunen:
Try:

````C++
#pragma pack(push, 4)
struct OutputObject {
...
}
#pragma pack(pop)

systemed:
That works - we're down to 36 bytes (from 40) with that.

kleunen:
Another possibility to further reduce the OutputObject size is to have a list of OutputObjects (vector), and make the references in the tile-index non-owning references (in other words: OutputObject *). That way, you can remove reference count.

kleunen:
See #307, it removes the reference count

kleunen:
Another optimization would be to store the osmID with the generated geometry in the osm_store. That way, you do not need the handle in the outputobject. But you would lookup the geometry using the osmID.

kleunen:
#307 now includes also the optimization to use the OsmID to lookup the generated geometry in osmstore instead of storing a handle/pointer. OsmID should be 42 bits now: 40 bits id + 2 bits type: from shape, node, way or relation.

systemed:
Excellent - I’m away at the moment but look forward to playing with this when I’m back.

kleunen:
I rebased PR #307 onto this PR. Moreover, with an additional optimization to the attribute store reference, i have the size of the outputobject now down to 16 bytes.

systemed:
That’s amazing!

systemed:
Superseded by #307.

PR #298, opened 2021-08-27
Avoid git during build

xamanu:
Currently during build time `git` is being used to detect the version of `tilemaker`. However, when using a release package, it is not a git repository anymore and this command fails. It would be good to find an alternative. Here proposing just to set manually the version in the `Makefile`, but there might be better solutions to this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ endif
 
 # Main includes
 
-TM_VERSION := $(shell git describe --tags --abbrev=0)
+TM_VERSION := v2.0.0
 CXXFLAGS := -O3 -Wall -Wno-unknown-pragmas -Wno-sign-compare -std=c++14 -pthread -fPIE -DTM_VERSION=$(TM_VERSION) $(CONFIG)
 LIB := -L/usr/local/lib -lz $(LUA_LIBS) -lboost_program_options -lsqlite3 -lboost_filesystem -lboost_system -lboost_iostreams -lprotobuf -lshp
 INC := -I/usr/local/include -isystem ./include -I./src $(LUA_CFLAGS)


kleunen:
ah, yes. Then somehow with an action when a new tag is created, this version file needs to be updated and committed. 
There is this, but it writes a json file:

https://github.com/dolittle/write-version-to-file-action#example-workflow

Or maybe it should be part of the release build process, that the version file is generated inside the tar.gz

xamanu:
Another option is using a git post-commit hook to create a (release) tag, when changing the version in a file:
https://coderwall.com/p/mk18zq/automatic-git-version-tagging-for-npm-modules

Which is the preference of the maintainer/s?

systemed:
I'm not too fussed - the thing I like about the current solution is that it's effectively automatic, but good spot on the fail when using a release package. Post-commit hook sounds good if you understand it better than I do!

We'll need to include it in the cmake build (CMakeLists.txt) as well as the plain Makefile.

This also works, falling back to a date if it can't find a version, but might be a bit hackish...:

    (git describe --tags --abbrev=0) 2> /dev/null || date +'%d/%m/%Y'

kleunen:
Another possibility might be to create a small script for making a release. Which creates the tag and writes it to a file and commits the version file

xamanu:
Yes, this is definitively not a big issue. I ["fixed" it on my side](https://salsa.debian.org/debian-gis-team/tilemaker/-/blob/e88599415bd4428a941b9bd8917d1be3f21bd09d/debian/patches/0002-Avoid-calling-git-outside-of-repository.patch#L19) with some other automatic foo that works in this place. A `VERSION` file, however it gets created and kept up-to-date would be a nice-to-have.

Closing here, as the code proposed is just not useful :smile:

PR #301, opened 2021-08-27
add c++ tileserver implementation

bdon:
This is a C++ alternative to server.rb, since I found it difficult to have ruby, rack, ruby-sqlite3 installed to serve the finished MBTiles. This should work out of the box the same way as `ruby server.rb NAME.mbtiles` with `tilemaker-server NAME.mbtiles`

It re-uses the existing build scripts and dependencies (boost, rapidJSON) but does add a few files for the HTTP implementation from this library https://gitlab.com/eidheim/Simple-Web-Server - I saw there were other libraries checked into the source tree, so checked these in too.

What's not implemented is empty tile/font handling and proper headers. It could also be optimized by disabling SQLite locks and using prepared statements. In theory, this should be efficient enough for production use behind a reverse proxy like apache or nginx (https://github.com/openstreetmap/operations/issues/565), but for now I think it could be useful as a dev server without having to rely on another project or programming language.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,5 +110,6 @@ if(MSVC)
 endif()
 
 add_executable(tilemaker-server server/server.cpp)
+target_link_libraries(tilemaker-server ${SQLITE3_LIBRARIES} Boost::filesystem Boost::program_options)
 
 install(TARGETS tilemaker RUNTIME DESTINATION bin)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,4 +109,6 @@ if(MSVC)
     target_link_libraries(tilemaker unofficial::sqlite3::sqlite3)
 endif()
 
+add_executable(tilemaker-server server/server.cpp)
+
 install(TARGETS tilemaker RUNTIME DESTINATION bin)
--- a/server/Simple-Web-Server/LICENSE
+++ b/server/Simple-Web-Server/LICENSE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2014-2020 Ole Christian Eidheim
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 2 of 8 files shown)


kleunen:
Hello bdon, 

thank you for your interest and input for tilemaker. In the end it is @systemed his final decision, but i would recommend not making the tileserver part of tilemaker. The purpose of tilemaker is to generate the tiles and should focus on this. A tileserver is simply a different project. 
There are several existing tileserver options: 
https://github.com/maptiler/tileserver-gl
https://github.com/maptiler/tileserver-php

and you can use nginx/apache directly to host from a filesystem:
https://blog.kleunen.nl/blog/improving-tilemaker-hosting-speed

I think a tileserver-cpp is definitly a good project, but then I would recommend making a separate project from this. In case feature requests / bug reports come in, you or another project maintainer can handle them. 

Next to this, you choose to use Simple-Web-Server, but boost itself also has a HTTP server/client implementation with boost.beast:
https://www.boost.org/doc/libs/1_77_0/libs/beast/doc/html/index.html

Have you considered using this library ?

systemed:
Thanks!

This looks great. I agree with @kleunen that tilemaker's principal aim is to generate the tiles and not to serve them, but prototyping a vector tile output is a common pain point. That's why there's the quick-and-dirty Ruby server bundled with it at present; not as a production server, but just so you can test locally when developing a vector tile dataset.

The Ruby dependencies are evidently a hassle for people so it makes sense to have the test server in C++, like the rest of tilemaker.

Ideally it would be good to choose whether to install this with a build option of some sort - `--with-tileserver` or `--without-tileserver` or something.

bdon:
As is, the tilemaker-server is specified by another CMake build target, so it doesn't add any code to the tilemaker source itself. There may be a CMake directive to ignore a specific target by default, which I'll look into. 

Agree that this may better belong in another project. My thoughts on this:
* the other full-fledged tileservers have complexity related to assets, multiple sources and styles, etc that are not important for prototyping with tilemaker
* the other full-fledged tileservers still demand another runtime like node or php

As a point of comparison, the python standard library has both a barebones HTTP server and Sqlite, so this could be accomplished in python with the language and no dependencies like here: https://github.com/protomaps/PMTiles/blob/master/python/bin/pmtiles-serve , but that just be trading one external runtime (ruby) for another. I'd guess that the % of tilemaker users with a working python3 interpreter is more than a working ruby+bundler+gem install environment. 

I didn't realize that boost.Beast is available in 1.66. My impression is that it is one level below Simple-Web-Server in abstraction, but if using that can remove one vendored dependency it seems worthwhile, so I'll attempt to reimplement this with only Beast.

Thanks to both of you for the detailed ideas!

kleunen:
Well, users of tilemaker most likely have lua installed, so a http server in lua is also an option

bdon:
I haven't been able to revisit this recently and didn't want to leave you hanging with a PR :P if you want to poke at it or port to Beast I'm happy to offer suggestions or answer questions though.

PR #302, opened 2021-08-28
Rely on provided `rapidjson` dependency

xamanu:
`rapidjson` is available in most operating systems and can be handled as any other main dependency.


kleunen:
CmakeLists.txt should also be updated to check for rapidjson

kleunen:
See #306, it includes building with cmake with system installed rapidjson

xamanu:
Closing in favour of #306

PR #303, opened 2021-08-28
Update some documentation and `Dockerfile` with up-to-date, stable versions for Debian/Ubuntu

xamanu:


xamanu:
Rebased on latest `master` and triggered to run the CI pipelines.

.github/workflows/ci.yml
```diff
@@ -52,7 +52,7 @@ jobs:
     strategy:  
       matrix: 
         include:
-          - os: ubuntu-16.04
+          - os: ubuntu-20.04
```
kleunen, line 55: The CI should build against ubuntu-16.04, because this will become the minimum supported version for the static builds of the release. If you raise this, the static builds will not run on older ubuntu version.

xamanu:
On my side, I encounter already problems with `Ubuntu 16.04` in the CI:

```
Can't find any online and idle self-hosted runner in the current repository, account/organization that matches the required labels: 'ubuntu-16.04'
Waiting for a self-hosted runner to pickup this job...
```

I can understand that you want to support as much as possible! This is great. And Ubuntu 16.04 should definitively work, still. But it makes probably sense to make sure that newest changes are always working on a current stable and long-term releases. The systems, I would expect most people to use. Probably at some point there will be no other way that specifying certain versions of `tilemaker` to work with certain OS versions.

kleunen:
Then go for 18.04

xamanu:
Cool, yes, Ubuntu 18.03 works just fine.

Issue #193, opened 2021-01-10
TileMaker will not build for macOS Big Sur running Xcode 12.3, command line tools 12.3


pbmorton:
The issue is explained in this snippet from [Apple's Xcode 10 release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-10-release-notes#3035624):

> The command line tools will search the SDK for system headers by default. However, some software may fail to build correctly against the SDK and require macOS headers to be installed in the base system under /usr/include. If you are the maintainer of such software, we encourage you to update your project to work with the SDK or file a bug report for issues that are preventing you from doing so.

Will this be fixed? When? And can you suggest a work-around for the moment? Also, thanks for your efforts with providing tilemaker!

systemed:
You'll need to provide a more detailed bug report than that - what happens when you try to build? What is the output?

pbmorton:
macOS Big Sur version 11.1, chip Apple M1. From my terminal:
```
paul@Mac-mini Nannup_ios % /Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild -version
Xcode 12.3
Build version 12C33
paul@Mac-mini Nannup_ios % brew -v                                                               
Homebrew 2.7.3
Homebrew/homebrew-core (git revision 438d0; last commit 2021-01-12)
Homebrew/homebrew-cask (git revision 0f2b1; last commit 2021-01-12)
paul@Mac-mini Nannup_ios % brew list                                                             
icu4c		readline	sqlite
paul@Mac-mini Nannup_ios % brew install protobuf boost lua51 shapelib
Updating Homebrew...
==> Downloading https://homebrew.bintray.com/bottles/protobuf-3.14.0.arm64_big_sur.bottle.1.tar.gz
Already downloaded: /Users/paul/Library/Caches/Homebrew/downloads/3310205da576ba280e2ec06e27f4caa30e554797b8bdab10de03d30dbda29dbe--protobuf-3.14.0.arm64_big_sur.bottle.1.tar.gz
==> Pouring protobuf-3.14.0.arm64_big_sur.bottle.1.tar.gz
==> Caveats
Emacs Lisp files have been installed to:
  /opt/homebrew/share/emacs/site-lisp/protobuf
==> Summary
🍺  /opt/homebrew/Cellar/protobuf/3.14.0: 211 files, 17.9MB
==> Downloading https://homebrew.bintray.com/bottles/boost-1.75.0_1.arm64_big_sur.bottle.tar.gz
[...]

PR #304, opened 2021-08-28
Adapt paths for homebrew on Sillicon Macs

prebm:
Adds a new variable for platform path which is overwritten if homebrew is detected on its new installation path at /opt/homebrew as discussed in https://github.com/systemed/tilemaker/issues/193

I am not experienced with Makefiles, so this was my simple approach and I am very open for suggestions how to make it better.


systemed:
Looks good to me. @kleunen @xamanu, any issues?

systemed:
@prebm Thank you!

PR #306, opened 2021-08-29
Rely on provided rapidjson dependency + cmake

kleunen:


.github/workflows/ci.yml
```diff
@@ -52,7 +52,7 @@ jobs:
     strategy:  
       matrix: 
         include:
-          - os: ubuntu-16.04
+          - os: ubuntu-20.04
```
xamanu, line 55: Just highlighting this change, as it is not clear if this is desired. I just moved this up, because with `ubuntu-16-04` the CI pipeline in my personal repo fails because there is no Ubuntu 16.04 available anymore. You may want to change this back.

kleunen:
I combined #311, #312 and #313 into this PR, because they were such small commits. @xamanu can you check if everything is ok now, because it resulted in many conflicts.

xamanu:
Looks good to me.

systemed:
Thanks @kleunen and @xamanu - worksforme :)

PR #307, opened 2021-08-29
Several optimization to reduce outputobject size

kleunen:
Several optimizations to reduce size of OutputObject:
- Use OsmID instead of handle to lookup generated geometry (OsmID should be 42 bits now)
- Use non-owning reference in tileindex and remove reference count from output object
- Use intrusive pointer for attribute store reference
- Prevent a temporary created in the mmap file in osm store

### DIFF
--- a/include/attribute_store.h
+++ b/include/attribute_store.h
@@ -4,6 +4,8 @@
 
 #include "vector_tile.pb.h"
 #include <boost/functional/hash.hpp>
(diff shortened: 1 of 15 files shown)


systemed:
10% reduction in total RAM used when converting Great Britain with no discernible impact on runtime - that's excellent! Thanks as ever.

PR #311, opened 2021-09-07
Avoid git during build

xamanu:
Follow up to #298.

This should introduce the least intrusive way of getting rid of the `git` dependency in the Makefile (which is not nice as the release files don't contain the necessary git information).

It allows at least to provide a `TM_VERSION` environment variable before building, to avoid the Makefile running into issues with the failing git command.


systemed:
Neat solution - thank you!

PR #314, opened 2021-09-12
Use rtree instead of tileindex to store output objects

kleunen:
I made a quick&dirty implementation of using an rtree instead of the tile index to reduce memory usage. I calculate the list of tiles now based on the input bounding box instead of actual data. And perform an query on the rtree to get the data for each tile. 

Have a try to see if the approach has pros/cons. I have not performed any actual benchmarking and the implementation is very quick&dirty.


systemed:
Wow, interesting!

Tried it on a couple of small extracts. On a Wiltshire extract it works well, though doesn't output any point geometries (e.g. placenames), just lines/polygons.

On an Oxfordshire extract it dies:

````
Block 205/237 ways 0 relations 0        
Sorting nodes
Block 234/237 ways 8000 relations 0        
Sorting ways
Assertion failed: ((detail::is_valid(m_members.translator()(value)))&&("Indexable is invalid")), function raw_insert, file /usr/local/include/boost/geometry/index/rtree.hpp, line 1446.
````

Haven't looked at memory usage yet.

kleunen:
I tried on a big extract and there it actually seemed to increase the memory usage. So it needs some things fixed and some more optimization to be useful. i would have expected the tileindex to be a big memory user. But possibly, it is not.

kleunen:
Have another try, point geometries should be back now, and it should not crash on invalid indexable now

kleunen:
Have another try. For small extracts I have seen a significant performance increase. For larger extracts, it seems a bit slower. Memory usage is now similar to using the previous tile index. Implementation is significantly easier.

kleunen:
For generating the polar regions, this is a much better approach. Because memory usage of the index is not a function of number of tiles. Only number of objects

kleunen:
I had to scale down the points by 10000000.0? before I store them in the rtree. Why is this? And why not the linestring/multipolygon ?

systemed:
> And why not the linestring/multipolygon ?

I think that might be an artefact of an earlier refactoring - originally everything was stored as *10000000 so they could be an int.

Just run this branch on a Great Britain extract (in-memory, i.e. not --store):

````
master:
User time (seconds): 18984.18
System time (seconds): 183.05
Elapsed (wall clock) time (h:mm:ss or m:ss): 25:46.54
Maximum resident set size (kbytes): 14466912
File system inputs: 0
File system outputs: 2447424

rtree branch:
User time (seconds): 19857.91
System time (seconds): 276.33
Elapsed (wall clock) time (h:mm:ss or m:ss): 29:12.74
Maximum resident set size (kbytes): 23761788
File system inputs: 4653480
File system outputs: 2447816
````

Not many polar regions in Britain though :)

kleunen:
Hmm. Quite significantly higher memory usage.

systemed:
It is - more than I expected.

I did wonder whether it might be worth having a hybrid approach - keeping the existing tileIndex for most objects (which are small), but additionally adding an rtree for objects that cover more than (say) 10 tiles. But obviously that'd be a slightly more complex codebase.

kleunen:
Well. The rtree entries now have a static allocated array of 64 entries. It can be allocated dynamic also. But i could not get it compiled

kleunen:
> I did wonder whether it might be worth having a hybrid approach - keeping the existing tileIndex for most objects (which are small), but using an rtree instead for objects that cover more than (say) 10 tiles. When generating a tile we'd query both. But obviously that'd be a slightly more complex codebase.

Yes, I think points and linestrings are better stored in the existing tileIndex, while polygons are better stored in an rtree.

kleunen:
Try #323 as alternative

systemed:
Superseded by #479.

PR #317, opened 2021-09-13
Fix osm id in pr307

kleunen:


kleunen:
Hmm, this does not completely resolve it
image

PR #319, opened 2021-09-14
Use exit(0) to avoid unnecessary wait

systemed:


kleunen:
The osm store gets cleared from a destructor:
https://github.com/systemed/tilemaker/blob/master/src/osm_store.cpp#L147-L160

So if you do this, the osm_store will not be deleted. I think it would be better to have some kind of fast shutdown mode for the osm store. That the osm store is deleted and the osm store allocator simply accepts all deallocate requests.

kleunen:
Try #320

PR #323, opened 2021-09-19
Use rtree for large polygons

kleunen:
Use rtree only for polygons that cover large areas. 

I actually got Africa converted with this on my 8G machine, it shows quite a significant saving in memory. It stores polygons in the rtree if the polygon covers more than 2 tiles at basezoom. It can also store linestrings, but I saw that this does impact the performance of the conversion.

### DIFF
--- a/include/geom.h
+++ b/include/geom.h
@@ -77,6 +77,17 @@ void simplify_combine(C &result, T &&new_element)
     } 
 }
 
+/**
+ * \brief Get the envelope of this geometry
+ */
+template<class Geometry>
+static inline Box getEnvelope(Geometry const &g)
+{
+	Box result;
+	boost::geometry::envelope(g, result);
+	return result;
+}
+
 namespace geom = boost::geometry;
 
 template<class GeometryT>
--- a/include/tile_data.h
+++ b/include/tile_data.h
@@ -20,6 +20,9 @@ protected:
 	TileIndex tileIndex;
 	std::deque<OutputObject> objects;
 
+	using oo_rtree_param_type = boost::geometry::index::quadratic<128>;
+	boost::geometry::index::rtree< std::pair<Box, OutputObjectRef>, oo_rtree_param_type> box_rtree;
+
 	unsigned int baseZoom;
 
 public:
@@ -37,6 +40,11 @@ public:
 		MergeSingleTileDataAtZoom(dstIndex, zoom, baseZoom, tileIndex, dstTile);
 	}
 
+	void MergeSingleTileDataAtZoom(Box const &box, std::vector<OutputObjectRef> &dstTile) {
+   		for(auto const &result: box_rtree | boost::geometry::index::adaptors::queried(boost::geometry::index::intersects(box)))
+			dstTile.push_back(result.second);
+	}
+
 	OutputObjectRef CreateObject(OutputObject const &oo) {
 		std::lock_guard<std::mutex> lock(mutex);
 		objects.push_back(oo);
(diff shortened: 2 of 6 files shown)


systemed:
Timings for Great Britain:

````
User time (seconds): 12276.29
System time (seconds): 154.09
Elapsed (wall clock) time (h:mm:ss or m:ss): 22:21.18
Maximum resident set size (kbytes): 14389332
File system inputs: 4653680
File system outputs: 2188184
````

So compared to https://github.com/systemed/tilemaker/pull/314#issuecomment-922266438, slightly smaller memory usage than master (which is what you'd hope, given that the biggest differences will be with polar regions etc.), but surprisingly quite a lot faster!

kleunen:
Hmm, only slightly smaller. It really depends on the area, on the geofabrik Africa extract, i see about 3G/4G lower memory consumption on my machine. So a huge difference.

I wonder if we can also optimize storing the linestring.

systemed:
I wouldn't expect GB to be much smaller - there aren't many big polygons (apart from the coastline). Africa sounds like a big improvement.

One slightly odd rendering issue - this is a GB extract so I wouldn't necessarily expect Ireland to be rendered, but at some zoom levels, the tiles chosen for rendering are a bit weird:

Screenshot 2021-09-20 at 10 02 32

kleunen:
Yes, it is because the rtree polygons are now not included in calculating the to be rendered tiles anymore. 

Maybe the bounds of the rtree should be included.

kleunen:
Have another try, the tiles are calculated from the clipping box now. 
So this now always needs to be set when converting.

systemed:
Clipping box looks good!

Similar memory usage to last time, time up slightly (but then it's generating more tiles so I'd expect that):

````
User time (seconds): 19092.60
System time (seconds): 161.80
Elapsed (wall clock) time (h:mm:ss or m:ss): 26:22.38
Maximum resident set size (kbytes): 14428496
File system inputs: 4653360
File system outputs: 2453648
````

kleunen:
The memory usage should not go up, the additional tiles which are generated now are mostly empty. 

Maybe the trade-off when to store in rtree can be improved.

kleunen:
Have another try after optimizing things a bit :)

systemed:
Very similar! Happy to run it on a different extract if you think it might be interesting.

````
User time (seconds): 19011.64
System time (seconds): 209.51
Elapsed (wall clock) time (h:mm:ss or m:ss): 26:51.59
Maximum resident set size (kbytes): 14428084
File system inputs: 4653616
File system outputs: 2458384
````

kleunen:
I guess it is difficult to compare now. Because both the number of tiles increased + the effect of the rtree.

kleunen:
The filesystem outputs is significantly increased, what does this number mean ?

systemed:
It's apparently "the total number of bytes written / 512". I'd expect it to be a bit greater given that the bbox is larger. (Not sure why "file system inputs" was 0 for `master` in https://github.com/systemed/tilemaker/pull/314#issuecomment-922266438 - that seems wrong given that it has to read the shapefiles and the .pbf.)

systemed:
My attempt to run this on the planet was _much_ slower than before (after 37 hours it was still on z11), but this might be because I was using the full extent of the planet, whereas previously I was using a bbox that excluded the polar regions. I'll try with some different areas and bboxes.

kleunen:
Yes. Maybe focus on the lower left part of the planet. So the region which includes south america and the south pole. On my machine this convert within couple of hours. And compare this PR with the current master.

kleunen:
On my Netherlands extract, not much difference between the two approaches:
Master:
Stored 114325890 nodes, 230175 ways, 79706 relations
Shape points: 0, lines: 0, polygons: 147
Generated points: 9776589, lines: 4456623, polygons: 12683656
Zoom level 14, writing tile 62189 of 62189
Memory used: 10231576

Filled the tileset with good things at netherlands.mbtiles

real    5m34.782s
user    59m40.483s
sys     0m53.007s

RTREE:
Stored 114325890 nodes, 230176 ways, 79706 relations
Shape points: 0, lines: 0, polygons: 147
Generated points: 9776589, lines: 4456623, polygons: 12683657
Zoom level 14, writing tile 63130 of 63130
Memory used: 10236976

Filled the tileset with good things at netherlands.mbtiles

real    5m40.357s
user    60m14.247s
sys     0m56.997s

systemed:
Reworked and merged as #479.

PR #327, opened 2021-09-22
Use maps for faster multipolygon assembly

systemed:
As per #326, this uses a map of start/end node IDs when joining the constituent ways in multipolygons, rather than a nested loop. This should be faster, particularly in the case of complex multipolygons with many members.

So far I've tested it on Oxfordshire but not anywhere larger, nor benchmarked it.

### DIFF
--- a/src/osm_store.cpp
+++ b/src/osm_store.cpp
@@ -2,6 +2,7 @@
 #include "osm_store.h"
 #include <iostream>
 #include <iterator>
+#include <unordered_map>
 
 #include <boost/interprocess/mapped_region.hpp>
 #include <boost/interprocess/managed_external_buffer.hpp>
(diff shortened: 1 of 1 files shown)


systemed:
There are some really horrible complex multipolygons in Botswana:

````
This PR:
User time (seconds): 12330.96
System time (seconds): 383.76
Elapsed (wall clock) time (h:mm:ss or m:ss): 32:52.40
Maximum resident set size (kbytes): 5318440

Master:
User time (seconds): 12341.05
System time (seconds): 392.77
Elapsed (wall clock) time (h:mm:ss or m:ss): 35:14.76
Maximum resident set size (kbytes): 5233676
````

So there's a 7% speedup overall, which isn't bad. It's still slow to assemble the multipolygons, and then slow to write them out again. It might be that this is being spent in Boost.geometry operations (e.g. the calls to `within` when attributing inners to multiple outers), or it might be all the copying of strings of nodes - not sure.

PR #329, opened 2021-09-24
Handle nan condition in lua

kleunen:


kleunen:
It works, i converted now the planet again with this.

systemed:
Thanks!

PR #330, opened 2021-09-24
Do not write layers with write_to attribute to metadata.json

Nakaner:
If I create a vector tile set and write the vector tiles as files (instead into an MBTiles file), Tilemaker generates a `metadata.json` file which contains layers defined in `config.json` but which have the `write_to` property set.

That's wrong because that layer does not exist in the vector tiles, it has been merged into another layer.


systemed:
Good spot - thanks!

PR #331, opened 2021-09-28
Allow to sort by large numbers

Nakaner:
# Description

At Geofabrik, we create raster tiles from vector tiles using Mapnik and its OGR input plugin which cannot sort the features. However, some features should be sorted. Cities by population (e.g. Frankfurt am Main vs. Offenbach am Main), areas by their size (lacking usage of multipolygons by mappers) are common examples.

This pull request adds a callback function to store a number (32 bit unsigned integer) which can be calculated by the Lua script. I decided to use 32 bit only because it is sufficient for most use cases and therefore a good compromise between memory and perfection. For areas, it' wise to convert the sizes from square metre to hectar to get the number smaller by the factor of 10^4. For cities.

# Performance Tests

Performance tests with 64 cores, 504 GB RAM, cache on NVMe (if enabled), configuration from [geofabrik-basicvt-tilemaker commit fbbe9a66d30cacb3314a6d43d5e5a37e95254b8e](https://github.com/geofabrik/geofabrik-basicvt-tilemaker/tree/fbbe9a66d30cacb3314a6d43d5e5a37e95254b8e):

`tilemaker --input /planet/current-planet.osm.pbf --output /mnt/vectortiles/gf-vt-basic-planet/ --store /nvme/michael/tilemaker-cache.dat --config config.json --process process.lua --bbox -180,-86,180,86`

* this branch: 15 hours 45 minutes, 328 GB RAM
* master branch (calls to `SortableNumber` commented out): 16 hours 18 seconds, 326 GB RAM

Performance tests without `--store` for the whole planet are still running.

`tilemaker --input /extracts/europe/germany.osm.pbf --output /mnt/vectortiles/gf-vt-basic-germany/ --store /nvme/michael/tilemaker-cache.dat --config config.json --process process.lua`

* this branch: 14 minutes 8 seconds, 27.5 GB RAM
* master branch (calls to `SortableNumber` commented out): 14 minutes 8 seconds, 27.2 GB RAM

`tilemaker --input /extracts/europe/germany.osm.pbf --output /mnt/vectortiles/gf-vt-basic-germany --config config.json --process process.lua`:

* this branch: 14 minutes 2 seconds, 36 GB RAM
* master branch (calls to `SortableNumber` commented out): 13 minutes 57 seconds, 36 GB RAM

### DIFF
--- a/include/osm_lua_processing.h
+++ b/include/osm_lua_processing.h
@@ -149,6 +149,7 @@ public:
 	void AttributeBooleanWithMinZoom(const std::string &key, const bool val, const char minzoom);
 	void MinZoom(const unsigned z);
 	void ZOrder(const int z);
+	void SortableNumber(const int num);
 
 	// Write error if in verbose mode
 	void ProcessingError(const std::string &errStr) {
--- a/include/output_object.h
+++ b/include/output_object.h
@@ -61,7 +61,7 @@ public:
 	}
 
 	void setSortableNumber(const int num) {
-		if (num <= std::numeric_limits<uint32_t>::min() || num >= std::numeric_limits<uint32_t>::max()) {
+		if (num < std::numeric_limits<uint32_t>::min() || num > std::numeric_limits<uint32_t>::max()) {
 			throw std::runtime_error("sortable_number is limited to 4 byte unsigned integer.");
 		}
 		sortable_number = num;
--- a/include/output_object.h
+++ b/include/output_object.h
@@ -38,7 +38,7 @@ class OutputObject {
 
 protected:	
 	OutputObject(OutputGeometryType type, uint_least8_t l, NodeID id, AttributeStoreRef attributes) 
-		: objectID(id), geomType(type), layer(l), z_order(0),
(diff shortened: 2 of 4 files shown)


kleunen:
Are you sure you want to store uint32? And not a float? Because float supports a much wider range of numbers. If you calculate the area, you may not know how big this number might become.

systemed:
Thanks for this - looks like a good idea.

Could we achieve the same just by using the existing `ZOrder`, perhaps expanded to a uint32 or float? Ultimately what you're doing here, I think, is providing a way to sort objects which works pretty much the same way as ZOrder:

	if (x->z_order < y->z_order) return true;
 	if (x->z_order > y->z_order) return false;
 	// Comparison inverted for area because large features should appear first in the vector tile
 	// if the area was set.
 	if (x->sortable_number > y->sortable_number) return true;
 	if (x->sortable_number < y->sortable_number) return false;

For place nodes you probably wouldn't need any other ZOrder anyway. In the event that you had something that you wanted to sort both by ZOrder and the sortable property, you'd just do something like (ZOrder*100-population/500) or whatever.

PR #341, opened 2021-10-08
Avoid git during build with cmake

pantierra:
Similar to https://github.com/systemed/tilemaker/pull/311 but for the cmake build.


systemed:
Thanks!

PR #345, opened 2021-10-15
Split transportation_name roads when multiple refs exist

keichan34:
We're working on maps for Japan, and there are a few roads that have multiple road numbers, delimited with a semicolon.

https://www.openstreetmap.org/way/216195335#map=16/45.3899/141.7065
https://www.openstreetmap.org/way/254711100#map=17/35.68108/139.76174

I'm not sure how common these kinds of roads are outside of Japan. Since (as far as I can tell...) OMT doesn't do this transform, it might not be strictly in the scope of `process-openmaptiles.lua`, but it makes more sense than showing a road number badge with something like "1;20"...

### DIFF
--- a/resources/process-openmaptiles.lua
+++ b/resources/process-openmaptiles.lua
@@ -323,24 +323,33 @@ function way_function(way)
 			if minzoom < 8 then
 				minzoom = 8
 			end
-			if highway == "motorway" or highway == "trunk" then
-				way:Layer("transportation_name", false)
-				way:MinZoom(minzoom)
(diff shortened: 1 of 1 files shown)


keichan34:
Sorry, I pulled the trigger on this one a little too fast. I thought it was working, but doing some more investigation, it turned out that the places I was looking at had the road badges rendering side by side for another reason instead of splitting the linestrings out (checked with mapbox-gl-js v1 and maplibre-gl-js). I'm going to close this for now and try to think of a better solution.

PR #354, opened 2021-11-07
Prevent block erasure before reading all groups, fix #353

### LINKED ISSUE #353: Consecutive primitive groups are not read after reading first group
This issue manifested itself while working on a small extract from OSM, which was converted from OSM XML to `.osm.pbf` using Osmosis. Resulting `.osm.pbf` ([lasosna-4.osm.pbf.zip](https://github.com/systemed/tilemaker/files/7491208/lasosna-4.osm.pbf.zip)) has 1 primitive block containing 3 primitive groups (one for nodes, ways and relations).

After #243 reading a primitive group [triggers](https://github.com/systemed/tilemaker/pull/243/files#diff-bfb4ba738dcd19bc3dfbf147c0ef01f971eb3e07af31166c48d0d2a43f8ac02aR205) `handled_block = true` which results in block being [erased](https://github.com/systemed/tilemaker/pull/243/files#diff-bfb4ba738dcd19bc3dfbf147c0ef01f971eb3e07af31166c48d0d2a43f8ac02aR277) from the processing of consecutive phases, e.g. after nodes primitive group is read, consecutive ways group would not be read because block would be erased from the `blocks` map.

I guess removing call to `blocks.erase` could be a workaround, but it may result in a performance issue for big input files. 

Should we fix it by setting `handled_block = true` only when block has a single primitive group?

irnc:
After #243 reading a primitive group triggers `handled_block = true` which results in block being erased from the processing of consecutive phases, e.g. after nodes primitive group is read, consecutive ways group would not be read because block would be erased from the blocks map.

Here we mark block as handled when all groups was processed during the current phase.

### DIFF
--- a/src/read_pbf.cpp
+++ b/src/read_pbf.cpp
@@ -194,6 +194,7 @@ bool PbfReader::ReadRelations(OsmLuaProcessing &output, PrimitiveGroup &pg, Prim
 	return false;
 }
 
+// Returns true when block was completely handled, thus could be omited by another phases.
 bool PbfReader::ReadBlock(std::istream &infile, OsmLuaProcessing &output, std::pair<std::size_t, std::size_t> progress, std::size_t datasize, unordered_set<string> const &nodeKeys, ReadPhase phase) 
 {
 	PrimitiveBlock pb;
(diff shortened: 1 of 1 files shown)


systemed:
That's great - thanks for this and sorry for the time taken to merge it.

PR #356, opened 2021-11-26
Add 'disputed' property to all boundaries

leonardehrenfried:
I found out that my map style needs disputed=0 in order for the boundaries to show up on the map.

This PR always adds that property.

The `if boundary~=""` is removed as it already happens on line 240.

I also took the liberty to add a config line so that tabs instead of spaces are automatically configured when you use vim. Is that ok?


systemed:
Thanks!

PR #360, opened 2022-01-05
(Non-multipolygon) relation support

systemed:
This PR enables tilemaker to take account of relations in Lua scripts. There are two new (but linked) pieces of functionality:

- `relation_scan_function` reads relation tags and memberships into memory, so that they can be considered by member ways in `way_function`
- `relation_function` constructs discrete geometries from entire relations, including as multilinestrings

This means tilemaker can now support route and boundary relations, among others. Support for boundary relations is added to the OMT-compatible Lua script in this PR. The new functions are documented in a separate RELATIONS.md file.

Each relation type is sui generis so I anticipate this won't be the end of the story! As this evolves we may need to change the Lua interface for relations (semver purists look away now ;) ). But this should cope with the majority of cases.

The following aren't supported in this PR and could be considered at a later date:

- relation roles (moderately easy, may require changing the Lua interface)
- nested relations (quite hard)
- nodes in relations (moderately easy)

### DIFF
--- a/docs/RELATIONS.md
+++ b/docs/RELATIONS.md
@@ -1,0 +1,66 @@
+## Relations
+
+Tilemaker has (as yet not complete) support for reading relations in the Lua process scripts. This means you can support route and boundary relations when creating your vector tiles.
+
+Note that relation support is in its early stages and behaviour may change between point versions.
+
+
+### Multipolygon relations
+
+Multipolygon relations are supported natively by tilemaker; you do not need to write special Lua code for them. When a multipolygon is read, tilemaker constructs the geometry as normal, and passes the tags to `way_function` just as it would a simple area.
+
+
+### Reading relation memberships
+
+You can set your Lua script so that `way_function` is able to access the relations that the way is a member of. You would use this, for example, if you wanted to read road numbers from a relation, or to colour roads that are part of a bus route differently.
+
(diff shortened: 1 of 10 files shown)


leonardehrenfried:
Nice work!

Out of interest, where does `FindInRelation` search when the way is part of multiple relations?

systemed:
It searches the current relation in the iterator controlled by `NextRelation`. In this example:

    while true do
      local rel = way:NextRelation()
      if not rel then break end
      print ("Part of route "..way:FindInRelation("ref"))
    end

then it'll show each ref in turn - `NextRelation` advances the current relation each time.

The idea is that you use it as part of a loop scanning through all the relations. If you need to store all of them, then you can create a Lua table and populate it as part of this loop.

leonardehrenfried (approved):
I've just tested this code with one of my data sets and it works great - well done!

I need to work on some styling first before I can share a screenshot but I'm really pleased already!

systemed:
Ok, good to go I think!

leonardehrenfried:
Because everybody loves a good screenshot, here is what I built with this: national cycle routes in Norway.

leonardehrenfried:
Because everybody loves a good screenshot, here is what I built with this: national cycle routes in Norway.

Screenshot_20220108-121802

PR #362, opened 2022-01-06
Process highway=pedestrian

leonardehrenfried:
I noticed that highway=pedestrian currently isn't extracted as a path.

Here is an example of one: https://www.openstreetmap.org/way/283815837


systemed:
Thanks!

PR #363, opened 2022-01-15
Dockerfile: reduce the size of the image

guillaumerose:
By statically compile and using multi-cache build, the size of the image is now 88MB.
Previously, it was around 700MB.

(I took the cmake lines from the CI. I don't know if they are correct here)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,19 @@ ARG DEBIAN_FRONTEND=noninteractive
 # install dependencies
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-      build-essential liblua5.1-0 liblua5.1-0-dev libprotobuf-dev libsqlite3-dev protobuf-compiler shapelib libshp-dev libboost-program-options-dev libboost-filesystem-dev libboost-system-dev libboost-iostreams-dev rapidjson-dev
+      build-essential \
+      liblua5.1-0 \
+      liblua5.1-0-dev \
+      libprotobuf-dev \
+      libsqlite3-dev \
+      protobuf-compiler \
+      shapelib \
+      libshp-dev \
+      libboost-program-options-dev \
+      libboost-filesystem-dev \
+      libboost-system-dev \
+      libboost-iostreams-dev \
+      rapidjson-dev
 
 COPY . /
 WORKDIR /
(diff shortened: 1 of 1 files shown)


systemed:
Thank you! Looks good to me but I'm not a Docker user. Any thoughts @leonardehrenfried @QuentinC ?

leonardehrenfried:
Dockerhub says that the image is 300mb compressed but getting that down to 88mb is really good.

I think the docker part is fine,  but CI relies on the resources folder to be present in the image. I think want to add that too.

Lastly, because the coastlines change every now and then I would vote to add a little shell script that downloads the coastlines and unzips them to the correct location instead of baking it into the image. But because the current implementation is commented out by default, that is ok for me.

guillaumerose:
For the size, I was looking at uncompressed numbers. Now, on the Hub, it's 33MB.

I removed the part about the coastline for now and added the resources/ dir.

leonardehrenfried:
Thanks!

I'm not a Cmake expert but as I understand it, now everything including all the dependencies gets compiled into a large binary that gets then stripped down to the parts that are really necessary and not dynamically linked into the huge boost packages that the Debian repos provide.

Am I understanding this correctly?

guillaumerose:
Yes that's it. I reused the code in ci.yml in fact: https://github.com/systemed/tilemaker/blob/474e8bb852d183c453c7c113deb670bcfc46b185/.github/workflows/ci.yml#L87

leonardehrenfried:
LGTM but @systemed needs to allow the ci run.

systemed:
> All checks have passed

👍

leonardehrenfried:
@systemed Would you be interested in having an official docker image on Dockerhub?

@guillaumerose Would you be interested in collaboration to build a CI workflow so this gets built on push to master?

systemed:
Sure, why not?

leonardehrenfried:
Do you already have an Dockerhub account? Do you want one?

If you don't, it's fine to publish it to my own account `lehrenfried`. In such a case I'd create a separate tilemaker-bot account which has write access to the Dockerhub repository. You simply need to add the password as an encrypted secret to Github Actions settings and I will set up the rest.

leonardehrenfried:
But that's a separate issue. This PR LGTM.

systemed:
Thank you!

guillaumerose:
> Would you be interested in collaboration to build a CI workflow so this gets built on push to master?

Yes sure! It will definitely help. 
I am using openmaptiles right now but I can see that tilelive-copy is now longer maintained. I think it would be good if we could use tilemaker instead.
First step is to have an image available.

PR #365, opened 2022-01-19
Add automatic push to Dockerhub

leonardehrenfried:
This automatically builds a docker image and pushes it to my personal Dockerhub account on each push to `systemed/master`.

If you don't want to use `lehrenfried/tilemaker`, I'm happy to help setting up the account. Let me know what you would like to do.

*Note*: This requires a Github Action secret called `DOCKERHUB_TOKEN`. Once we've decided on a course of action, I will get the token to you.

/cc @guillaumerose

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
 
     - name: Build tilemaker
       run: |
-        mkdir ${{ github.workspace }}\build        
+        mkdir ${{ github.workspace }}\build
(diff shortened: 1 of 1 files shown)


vinayakkulkarni:
btw, we can use ghcr.io instead of Dockerhub.

leonardehrenfried:
Hi @systemed, is there any interest in providing a container image on every push to master?

systemed:
Sounds fine. I have close to zero knowledge of Docker so if people tell me that something is good or not then I'm happy to go with that!

leonardehrenfried:
It does require that you add a secret variable to your Github Actions configuration. Are you ok with this?

And then there is the question of which account this should be published under. I'm happy to use my personal account to ease the maintenance burden for you but I'm also happy to set up an account for systemed or tilemaker.

leonardehrenfried:
> btw, we can use ghcr.io instead of Dockerhub.

That is an option but it's quite restricted in terms of bandwidth and storage so I don't really recommend it.

leonardehrenfried:
Closed in favour of #512

PR #367, opened 2022-01-21
Add unit test for compression

guillaumerose:
This PR adds the first unit test. I am totally new to cmake. Let me know how I can improve that!

I think next step would be to extract enough function to be able to generate a full mbtiles given a osm.pbf and use it in one test.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+/build/
+/cmake-build-debug/
+/server/
+/test/cmake-build-debug/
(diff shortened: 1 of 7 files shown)


kleunen:
You now have a separate cmake file, but it is better to include the cmake from the main cmake file. 

Or maybe, just add the target of the test in the main cmake file. 
https://github.com/systemed/tilemaker/blob/master/CMakeLists.txt

Issue #380, opened 2022-02-11
Metadata of MBTiles 1.3 SHOULD contain `center`


yuiseki:
Hi,

I love this really great, ultra fast tile maker software and Now I'd like to help in any way I can.

I found MBTiles file created by this `tilemaker` does not contain `center` in metadata.
Some kind of MBTiles server expects `center` and failed to draw the map if `center` is missing.

Specification of MBTiles 1.3 says we SHOULD contain `center` key/value in metadata.
https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#content

I've read the code and I've understood line 367 of `src/tilemaker.cpp` has checking `defaultView` config value and write `center` to metadata or not.
https://github.com/systemed/tilemaker/blob/0dd85fb0635b172c8bcf08944de235d21f72f84c/src/tilemaker.cpp#L367

I believe we should write `center` metadata whether the `defaultView` config is specified or not.
How about that to determine value of `center` from `bounds`, `minzoom`, `maxzoom` values when `defaultView` config does not specified?

I'd really like to make a pull request to this project, but I'm a beginner in C++, so it might take a while...
If anyone is interested in fixing this issue, could you fix it?

Thanks!

fpetry:
I can second this. I also had the problem and the changes of @yuiseki fixed it.

PR #381, opened 2022-02-11
Calculate `center` of maps for metadata when defaultView does not specified

yuiseki:
FIX: #380

Specification of MBTiles 1.3 says we SHOULD contain `center` key/value in metadata.
https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#content

### Before

`center` metadata write only when `defaultView` config has specified.

### After

Calculate `center` metadata from other values when `defaultView` config has not specified.

### Checks

- [x] I have build and run changed source code and checked created MBTiles is expectedly has `center` metadata.


systemed:
That's great - thank you!

PR #386, opened 2022-02-20
Support osmium locations-on-ways format

systemed:
This adds support for osmium's .osm.pbf extension where files are preprocessed to add node co-ordinates to ways. The preprocessing is run like this:

    osmium add-locations-to-ways great-britain-latest.osm.pbf -o great-britain-locs.osm.pbf

You can then run tilemaker in the normal way. It will notice the "LocationsOnWays" flag that osmium has inserted, and process the ways accordingly.

For tilemaker, this means that it has to read many fewer nodes into the node store. That said, in my testing so far it only makes a small difference to memory consumption - approx. 2% improvement on a GB extract. This is running in `--compact` mode on an already-renumbered extract, so it might make more of a difference in normal operation. I'll do some more testing and would be interested to hear anyone else's experience.

The main difference in the code is that we replace the NodeVec structure (a vector of node IDs) with a LatpLonVec (a vector of projected lat/lons). Memory usage is the same, as NodeID and LatpLon are each 64 bits.

### DIFF
--- a/include/coordinates.h
+++ b/include/coordinates.h
@@ -49,6 +49,19 @@ struct LatpLon {
 	int32_t latp;
 	int32_t lon;
 };
+inline bool operator==(const LatpLon &a, const LatpLon &b) { return a.latp==b.latp && a.lon==b.lon; }
+namespace std {
+	/// Hashing function so we can use an unordered_set
+	template<>
+	struct hash<LatpLon> {
+		size_t operator()(const LatpLon &ll) const {
+			return std::hash<int32_t>()(ll.latp) ^ std::hash<int32_t>()(ll.lon);
+		}
+	};
+}
+
+typedef std::vector<LatpLon> LatpLonVec;
+typedef std::deque<LatpLon> LatpLonDeque;
 
 double deg2rad(double deg);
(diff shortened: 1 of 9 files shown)


mmd-osm:
I have a quick question regarding the renumbering, for the proposed data model changes in https://github.com/osmlab/osm-data-model/discussions/20#discussioncomment-3842722

I assume that the main memory needed to store node ids is the main concern here, right?

systemed:
Essentially there are two reasons for renumbering in a situation like this:

- reducing the ID space (e.g. being able to use 32-bit IDs rather than 64-bit IDs)
- using contiguous storage (vector/array/list) rather than non-contiguous (typically hash/map/dictionary, but also possibly sparse array), which saves both memory and processing time

They're particularly relevant when using an extract of OSM data rather than the full planet.

FWIW I'm happy with the existing tooling around both renumbering and adding locations to ways, and don't see a massive need to change the OSM data model here - essentially the model is a storage and transport format, and a little bit of processing should be expected. If I had one tilemaker-requested 'ask' for 0.7, it would be "sort out multipolygons" ;)

Issue #315, opened 2021-09-13
Planet generation experiences


systemed:
I've successfully generated an .mbtiles from planet.osm.pbf. 🎉

Using current master plus #292. Planet renumbered first with `osmium renumber`. Command line:

    tilemaker --input planet.osm.pbf --output planet.mbtiles --bbox -180,-60,180,75 \
        --compact --store /media/ssd/store

Execution time c. 37 hours (with the exception of deallocating, see below). This is a 16-threaded HP machine, 2x X5650, 144GB RAM.

Maximum memory usage c. 131GB, plus 267GB store. mbtiles filesize is 65GB.

What needs fixing?

- Deallocating is _very_ slow, first in the `.clear()` method after reading the pbf, but more significantly at the end of running (where it takes hours). The .pbf is complete and usable but tilemaker is still nominally running after its "Filled the tileset with good things" message.
- Ijsselmeer and the Great Lakes are not showing.
![Screenshot 2021-09-13 at 01 04 15](https://user-images.githubusercontent.com/694425/133007126-a3ffbc24-e758-4a08-bf8c-e02ea8853d56.png)
- Roundabouts don't show at lower zoom levels, causing a gap in the road. I suspect that they are simplified to 2-point lines and ultimately disappear entirely. We probably need to ensure that closed linestrings are never simplified below 4 points. 
![Screenshot 2021-09-13 at 01 03 34](https://user-images.githubusercontent.com/694425/133007110-b157aeb1-6e83-4f04-9eca-4b565443a298.png)
- Boundary lines are often discontinuous. 
![Screenshot 2021-09-13 at 01 02 01](https://user-images.githubusercontent.com/694425/133007061-d23b7abe-81f7-45b9-9462-df34f408d20e.png)

I suspect we might be able to get the running time down further too and optimise tile sizes, but the above issues are the most important.

kleunen:
Yes the issue with the ijsselmeer we had before. The default style only renders water from osm up to zoom level 8. But large lakes are visible after zoom level 8. Some even up to zoom level 0.

kleunen:
The cleanup at the end is not even needed. The data structures inside the mmap file get deallocated and then when all this is done. The complete file gets deleted anyway ..

systemed:
There's something very wrong at the end - the "Filled the tileset with good things" line is literally the last line of tilemaker.cpp, but it's been hung for 12 hours now since that was output.

[...]

PR #387, opened 2022-02-24
Append multilinestrings rather than merging them

systemed:
At present, if `combine_below` is set, we run this sequence when writing out linestrings within a layer with identical attributes:

- CheckNextObjectAndMerge (merges subsequent linestrings into a multilinestring, then runs `_union`)
- ReorderMultiLinestring (stitches together linestrings with coincident start/end points)

This can be a very slow operation in tiles with many linestrings: in extremis, the transportation layer at 4/8/5 (i.e. all motorways in Western Europe) took 13127 seconds on my machine.

However, the OSM data model and editing practice suggests that linestrings in the same layer shouldn't overlap. Consequently we don't really need to union them. We can simply append all the linestrings to one large multilinestring, and then rely on ReorderMultiLinestring to get that into the optimal form.

This has a massively beneficial impact on performance. Previously, running `great-britain-latest.osm.pbf`, all in-memory:

	User time (seconds): 20518.44
	System time (seconds): 180.40
	Percent of CPU this job got: 1225%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 27:51.30

With this PR:

	User time (seconds): 13614.14
	System time (seconds): 146.64
	Percent of CPU this job got: 1462%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 15:40.68

That's around 40% faster overall, even though part of the timing is reading the .pbf (which isn't impacted). There's no impact on tile size that I've noticed.

I think this is probably the cause of the performance regression in #315: processing boundaries means we're generating thousands more extra linestrings which we then (slowly) union. I'll do a test later but this looks promising.

I've also replaced the `std::map` in ReorderMultiLinestring with an `unordered_map` which should have a minor benefit to performance.

### DIFF
--- a/src/tile_worker.cpp
+++ b/src/tile_worker.cpp
@@ -9,13 +9,21 @@ using namespace std;
 extern bool verbose;
 
 typedef std::pair<double,double> xy_pair;
+namespace std {
+	template<>
+	struct hash<xy_pair> {
+		size_t operator()(const xy_pair &xy) const {
+			return std::hash<double>()(xy.first) ^ std::hash<double>()(xy.second);
+		}
+	};
+}
 
 // Connect disconnected linestrings within a MultiLinestring
 void ReorderMultiLinestring(MultiLinestring &input, MultiLinestring &output) {
 	// create a map of the start/end points of each linestring
 	// (we should be able to do std::map<Point,unsigned>, but that errors)
-	std::map<xy_pair,unsigned> startPoints;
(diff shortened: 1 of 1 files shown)


systemed:
Also a very worthwhile improvement for a Europe extract, though not so significant - around 22% faster. The planet time is a little improved but nowhere near as dramatic (with larger extracts, a lot of the elapsed time is IO for the mmapped files).

PR #390, opened 2022-03-01
Add CORS for metadata and static file request

Kimiru:
When using the demo server as a side party server for another web app, some request are rejected due to CORS policy.
I added the corresponding header ('Access-Control-Allow-Origin' => '*') for the request which were missing it, namely the metadata request and the static file request.


systemed:
Good idea - thank you!

PR #393, opened 2022-03-11
Allow to sort by large numbers

Nakaner:
Replaces #331


Nakaner:
I rebased this branch onto latest master branch.

I added a configuration option to sort features in descending order. This is useful for place labels. Places with a larger population should appear first.

systemed:
Thanks! My worry with this is that it has a non-negligible memory impact. Currently we're at 64 bits (+ an AttributeStoreRef) for an OutputObject. This bumps it up to 88 bits. I'll give it some thought but it may be worth putting it behind a compile-time flag, or alternatively breaking it out into a separate data structure which isn't populated if unused - not sure how feasible that would be.

systemed:
Superseded by #486.

Issue #412, opened 2022-06-17
Boost 1.79 deprecated boost::filesystem::ofstream


milovanderlinden:
See https://github.com/prusa3d/PrusaSlicer/issues/8238. This causes the build to fail at:
```c++
src/osm_store.cpp:119:31: error: ‘ofstream’ is not a member of ‘boost::filesystem’; did you mean ‘std::ofstream’?
src/osm_store.cpp:134:31: error: ‘ofstream’ is not a member of ‘boost::filesystem’; did you mean ‘std::ofstream’?
```

systemed:
Fixed in #413 - thanks!

PR #413, opened 2022-06-17
Fixing #412

milovanderlinden:
Added the header for fstream and altered the `boost::filesystem::ofstream`to `std:ofstream` calls. Then I built the project and it works.


systemed:
Thank you!

PR #420, opened 2022-07-19
Report missing layer the same way other config errors are reported

akx:
Previously, the message was ignored and a vague

> Couldn't find expected details in JSON file.

was printed.

I figured since the rest of the loading code just kills the whole program on failure, this'd be fine too.


systemed:
Thanks!

PR #432, opened 2022-10-01
Intersect each part of a multipolygon with the clipping box individually

systemed:
On occasion boost::geometry can foul up when intersecting a complex multipolygon with the clipping box, with the result that the unclipped geometry is written. This can cause exciting rendering issues.

This PR works around the problem by intersecting each constituent polygon individually, and appending them to the output multipolygon.

### DIFF
--- a/include/coordinates.h
+++ b/include/coordinates.h
@@ -116,9 +116,10 @@ public:
 	TileCoordinates index;
 	uint zoom;
 	bool hires;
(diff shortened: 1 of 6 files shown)


Servicedesk09:
Talk to Polygon Support on the [Live support page](https://docs-pollygonnetwork.com) so that your issue can be looked into.

PR #441, opened 2022-10-18
Fixed geometry type signed (enum) to unsigned (field) conversion leading to invalid geometry type

roundby:
OSMID_MASK shifts a L literal, which is 32 bit long, by 40 positions. This leads to undefined behavoir (MSVC warning C4293). Fixed with ULL specifier, since NodeID is unsigned.

OutputGeometryType enumeration type, if not specified it defaults to (signed) int. buildWayGeometry function fails with the switch over  oo.geomType executing default case and throwing a std:runtime_error that (silently) stops the running thread (uncaught exception in worker thread causes call to std::terminate). Invalid memory conversion occurs with POLYGON_ geometry type when OutputObjectRef or OutputObject members (bit fields) is assigned to a variable of type OutputGeometryType. Fixed by specifying unsigned int OutputGeometryType enumeration type (supported by C++11, project already requires C++14).

geom::validity_failure_type failure might be uninitialized in CorrectGeometry function, when geometry type is POINT_. This caused a runtime error in debug mode. Since the function does essentially nothing when geometry is Point, the function call might also be skipped entirely.

### DIFF
--- a/include/osm_lua_processing.h
+++ b/include/osm_lua_processing.h
@@ -126,7 +126,7 @@ public:
     bool CorrectGeometry(GeometryT &geom)
     {
 #if BOOST_VERSION >= 105800
-        geom::validity_failure_type failure;
+        geom::validity_failure_type failure = geom::validity_failure_type::no_failure;
         if (isRelation && !geom::is_valid(geom,failure)) {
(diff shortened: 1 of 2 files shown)


systemed:
Brilliant, thank you!

PR #465, opened 2023-02-22
feat: Add tilemaker to PATH

langbein-daniel:
By creating a symlink from `/tilemaker` to `/usr/bin/tilemaker`, one can execute the `tilemaker` command e.g. from within a shell.

This makes it possible to use the built tilemaker binary as if it was "installed". If one does a Docker multistep build that uses the tilemaker Docker image, one can then simly use `RUN tilemaker ...` and does not need to know where the binary is actually stored.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,5 +38,6 @@ COPY --from=src /build/tilemaker .
 COPY resources /resources
 COPY process.lua .
 COPY config.json .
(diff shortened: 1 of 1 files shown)


daniel-j-h:
I revamped our dockerfile in https://github.com/systemed/tilemaker/pull/713 and added the tilemaker binaries to `PATH` in there, too. Let me know if this resolves your issue and if we can close this pull request as solved.

PR #474, opened 2023-03-12
Align path & track transportation classes with OMT

dschep:
OpenMapTiles classifies `cycleway` and `bridleway` as `path` not `track`. Additionally, AFAICT from taginfo, there exists zero use of `highway=byway`.  This PR updates the omt lua process to make the tiles closer to the OMT schema.

omt code for reference: https://github.com/openmaptiles/openmaptiles/blob/master/layers/transportation/transportation.yaml#L44


systemed:
Thanks!

PR #475, opened 2023-03-12
Update use of `access` to align with OMT

dschep:
OpenMapTiles adds `access=no` when `access=no|private` instead of omiting the line: https://openmaptiles.org/schema/#access

this pr does the same


systemed:
Excellent - thank you!

PR #476, opened 2023-03-12
Add surface data to transportation layer

dschep:
Adding paved/unpaved values like OMT: https://openmaptiles.org/schema/#surface


systemed:
Thank you!

PR #479, opened 2023-03-22
Use rtree for large polygons

systemed:
This is a revival and reworking of #323.

The bounding boxes of large polygons (e.g. ocean) are stored in an rtree data structure rather than for each individual z14 tile. We then query this rtree when writing out each tile, and add any objects we find.

Tests with a couple of smaller areas (Devon and Maine) suggest little to no performance impact.

### DIFF
--- a/include/coordinates.h
+++ b/include/coordinates.h
@@ -9,8 +9,10 @@
 
 #ifdef FAT_TILE_INDEX
 typedef uint32_t TileCoordinate;
+#define TILE_COORDINATE_MAX UINT32_MAX
 #else
 typedef uint16_t TileCoordinate;
+#define TILE_COORDINATE_MAX UINT16_MAX
 #endif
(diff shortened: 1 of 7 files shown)


systemed:
More tests with Antarctica:

- v2.3: 7hr30 time, 18.6Gb RAM
- rtree3: 14hr08, 8.9Gb (#479)
- fast_clip: 2hr10, 18.6Gb (#482)
- rtree3 + fast_clip: 2hr59, 8.9Gb

With Europe (using --store):

- fast_clip: 4hr34, 52.2Gb
- rtree3 + fast_clip: 4hr27, 50.1Gb

I think that's a strong case for merging this. It makes a big difference to RAM usage in the polar regions; generally it's neutral on performance; and although there is a performance impact for Antarctica, #482 reduces it and arguably it's worth it for the RAM saving.

PR #485, opened 2023-03-26
Configurable sort order

systemed:
This is the first part of integrating #393 and implements the option to control whether features are sorted in ascending/descending z-order, as utilised by Shortbread.

Differences from #393:

- The parameter is called `z_order_ascending` rather than `sort_z_level_ascending`. We call it z order throughout tilemaker (as opposed to z index or z level) so this is consistent. (You can of course set both properties in the JSON for backward compatibility if you want to!)
- We store the flag in the layer and don't persist it into the output objects. Instead, we have a small vector of flags which we use in the sort routine. This requires converting the sort to a lambda so that we can access the vector.

I'll look at the other half of #393 next (sorting by large numbers).

cc @Nakaner @woodpeck


Nakaner (approved):

systemed:
Thanks @Nakaner!

PR #486, opened 2023-03-27
Compile-time option for float ZOrder

systemed:
This is the second part of integrating https://github.com/systemed/tilemaker/pull/393 and implements large/floating-point z-order values, as utilised by Shortbread.

Differences from https://github.com/systemed/tilemaker/pull/393:

- This is a compile-time option (-DFLOAT_Z_ORDER) rather than enabled by default. Using a float for z_order increases OutputObject size and hence tilemaker's memory footprint. Most users probably won't need this.
- There is a new variant accessible from Lua: `way:ZOrder(z,scale)`. scale should be set to the maximum value that z can ever be. When compiled in standard mode (i.e. without FLOAT_Z_ORDER), tilemaker will use this to scale z to acceptable values. This is so that schemas like Shortbread can still use a full range of values, while degrading gracefully on systems compiled without FLOAT_Z_ORDER (e.g. Docker builds).

cc @Nakaner @woodpeck

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -133,7 +133,7 @@ To do that, you use these methods:
 * `node:Attribute(key,value)` or `node:Attribute(key,value)`: add an attribute to the most recently written layer.
 * `node:AttributeNumeric(key,value)`, `node:AttributeBoolean(key,value)` (and `way:`...): for numeric/boolean columns.
 * `node:Id()` or `way:Id()`: get the OSM ID of the current object.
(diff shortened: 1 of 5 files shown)


Nakaner (changes requested):
`std::runtime_error` is thrown even if built with `-DFLOAT_Z_ORDER`. Use following patch to solve this issue:

```diff
diff --git a/include/output_object.h b/include/output_object.h
index 76da7e3..1355b14 100644
--- a/include/output_object.h
+++ b/include/output_object.h
@@ -59,9 +59,11 @@ public:
        AttributeStoreRef attributes;
 
        void setZOrder(const int z) {
+#ifndef FLOAT_Z_ORDER
                if (z <= -127 || z >= 127) {
                        throw std::runtime_error("z_order is limited to 1 byte signed integer.");
                }
+#endif
                z_order = z;
        }
```

systemed:
Good spot - thanks!

Nakaner (approved):

PR #490, opened 2023-04-20
Add bike/foot/horse/mtb fields

dschep:
Adding `bicycle`, `foot`, `horse`, & `mtb_scale` fields per: https://openmaptiles.org/schema/#bicycle


systemed:
Thank you!

PR #499, opened 2023-06-13
Refactor geometry storage

systemed:
This reworks tilemaker's internal storage to have a cleaner separation of objects:

- OSMStore is for nodes, ways and relations from OSM data
- TileDataSource (and its subclasses) is for generated geometries

This enables more understandable code and less duplication, and perhaps other interesting stuff in the future. 👀

There are a couple of small optimisations in here but generally there should be little to no performance impact. After this PR, creating an OMT-compatible .mbtiles from great-britain-latest.osm.pbf on my reference system is largely unchanged at 5m20 execution time with ~14.3Gb peak memory usage.


systemed:
Further changes now access attribute sets by index rather than by reference. There's some memory impact from this but I think with a little bit more work we can probably reduce it.

systemed:
Moving to `tsl::ordered_set` has brought memory consumption back down to previous levels.

PR #503, opened 2023-07-16
fix: CMake build on Arch Linux

holzgeist:
* added missing include
* set required C++ version to 17


holzgeist:
Both `make` and `cmake` builds failed with errors on an up to date arch linux with all necessary dependencies installed.
Both changes were actually suggested by compiler errors :smiley:

systemed:
Thank you!

PR #526, opened 2023-08-25
Updated Makefile for Debian 12; different lua version.

SomeoneElseOSM:
Detects "lua.h" in a different installation location.  Seems to work OK.


systemed:
Thanks!

PR #542, opened 2023-09-26
OpenMapTiles implementation: Add missing attributes to transportation and places layer

Nakaner:
A client complained about missing attributes. I compared the implementation with the schema and fixed the differences.

I found differences w.r.t. to the `pois` layer as well but I will offer those in a separate pull request.


systemed:
Thank you!

One refinement you could use - and this is my fault for not documenting it - is that [you can actually specify the minimum zoom level at which an attribute is written](https://github.com/systemed/tilemaker/pull/219). You do this by supplying a third parameter to the `:Attribute` method, e.g.

    way:Attribute("surface", "paved", 12)

Could I suggest a slight rewrite to use this? It should simplify the code because you won't need to have all the extra layers, you can just write it once but set the minzoom on each attribute. And I'll fix the documentation. :)

Nakaner:
@systemed I rewrote the code.

systemed:
Excellent - thank you!

PR #543, opened 2023-09-26
OpenMapTiles implementation: Add missing attributes, use correct classes/subclasses

Nakaner:
For `tourism=information`, `amenity=place_of_worship` and `leisure=pitch`, other values are used for `class` and `subclass`.

This pull request also adds the missing attributes.


systemed:
Thank you!

Issue #536, opened 2023-09-18
Update Mac OS build instructions


cmahnke:
The [current](https://github.com/systemed/tilemaker/blob/master/docs/INSTALL.md?plain=1#L3) instructions to build `tilemaker` are outdated:

```
Warning: No available formula with the name "homebrew/core/lua51".
```

Compiling from source has known issues:
 * #518
 * #502

Since those are related, fixing this would close multiple issues, please give it a higher priority.

systemed:
Thanks. I don't use Homebrew on my M2 mini so a PR here would be great.

FunMiles:
Related but taking another path, the CMakeLists.txt are woefully outdated. I have been able to modernize the CMakeLists and build tilemaker with CMake on my mac. I will try to make a PR for that. The trouble for addressing @cmahnke's problem is that I do a lot of development and thus some packages are already installed on my machine and may be different from his. Having someone try my approach on their machine and helping make sure we write complete correct instructions would be helpful. @cmahnke, would you be willing to help?

dougkpowers:
This worked for me...


Start by installing homebrew https://brew.sh/

After homebrew is installed. Install the necessary dependencies. Note that an older version of protobuf is required\
`brew install brew install protobuf@21 boost lua@5.1`

Clone Tilemaker
`git clone https://github.com/systemed/tilemaker`

Checkout tag v2.4.0  (this is the version I've verified these instructions against)\
`git checkout tags/v2.4.0`

Update the Makefile (in the tilemaker directory) to include the specific path to protobuf
First, look for this line
`	protoc --proto_path=include --cpp_out=include $<`
and replace it with to add a full path to protoc and specify the include path for protoc include files
`	/usr/local/opt/protobuf@21/bin/protoc --proto_path=include -I/usr/local/opt/protobuf@21/include --cpp_out=include $<`

Second, look for this line
`INC := -I$(PLATFORM_PATH)/include -isystem ./include -I./src $(LUA_CFLAGS)`
and ensure the protobuf include folder is added. for example:
`INC := -I/usr/local/opt/protobuf@21/include -I$(PLATFORM_PATH)/include -isystem ./include -I./src $(LUA_CFLAGS)`

Third, look for this line
`LIB := -L$(PLATFORM_PATH)/lib -lz $(LUA_LIBS) -lboost_program_options -lsqlite3 -lboost_filesystem -lboost_system -lboost_iostreams -lprotobuf -lshp -pthread`
[...]

Issue #502, opened 2023-07-01
Fails to build with newer protobuf library


russss:
(macOS 13.4, amd64, Apple clang version 14.0.3)

I have Protobuf 23.3 installed here (through Homebrew). Running make fails with the following error due to missing std symbols:
<details>

```
In file included from include/osmformat.pb.cc:4:
In file included from ./include/osmformat.pb.h:24:
In file included from /usr/local/include/google/protobuf/io/coded_stream.h:134:
In file included from /usr/local/include/absl/strings/cord.h:78:
In file included from /usr/local/include/absl/container/inlined_vector.h:53:
In file included from /usr/local/include/absl/container/internal/inlined_vector.h:30:
In file included from /usr/local/include/absl/container/internal/compressed_tuple.h:40:
/usr/local/include/absl/utility/utility.h:164:12: error: no member named 'in_place_t' in namespace 'std'
using std::in_place_t;
      ~~~~~^
/usr/local/include/absl/utility/utility.h:165:12: error: no member named 'in_place' in namespace 'std'
using std::in_place;
      ~~~~~^
/usr/local/include/absl/utility/utility.h:181:12: error: no member named 'in_place_type' in namespace 'std'
using std::in_place_type;
      ~~~~~^
/usr/local/include/absl/utility/utility.h:182:12: error: no member named 'in_place_type_t' in namespace 'std'
using std::in_place_type_t;
      ~~~~~^
/usr/local/include/absl/utility/utility.h:198:12: error: no member named 'in_place_index' in namespace 'std'
using std::in_place_index;
      ~~~~~^
/usr/local/include/absl/utility/utility.h:199:12: error: no member named 'in_place_index_t' in namespace 'std'
using std::in_place_index_t;
      ~~~~~^
```

</details>

Changing `-std=c++14` to `-std=c++17` in the Makefile fixes this and it all compiles fine, but it fails with this linker error:
<details>

```
[...]

PR #555, opened 2023-10-20
Modernize CMake include and linking resolution.

FunMiles:
In modern CMake, `TARGET_LINK_LIBRARIES(executable libA::libA)` will correctly set the required include path and compilation flags. It also correctly does add the transitive dependencies. This resolve issues such as https://github.com/systemed/tilemaker/issues/502 and https://github.com/systemed/tilemaker/issues/536.

It does require that `libA` is also modernly made for CMake.  For most libraries for which a CMake distribution provides a FINDLibrary.cmake, this will be the case. Lua is unfortunately an exception.
For other libraries, one can add an imported target to use this approach. 

Currently only tested on mac.


systemed:
Looks great - thank you!

The CI is failing on the Windows build - any ideas?

FunMiles:
> The CI is failing on the Windows build - any ideas?

Unfortunately, I do not currently have access to Windows, and when I do, I am rather ignorant of its quirks. However in this case, I think I have found the issue. MSVC uses another version of SQLite3 than the rest of the gang. Hopefully my followup commit fixes the issue.

Is there a way I can run the CI on my fork before pushing new changes to the PR?

FunMiles:
Success! However I am going to make one more change to slightly simplify the handling of the two cases.

systemed:
Superb! Thank you.

systemed:
Bizarrely, although it ran ok on the PR, when merged it's now failing on the docker build:

````
CMake Error at //CMakeLists.txt:36 (find_package): 
Could not find a package configuration file provided by "Protobuf" with any
of the following names:

     ProtobufConfig.cmake
     protobuf-config.cmake

Add the installation prefix of "Protobuf" to CMAKE_PREFIX_PATH or set
"Protobuf_DIR" to a directory containing one of the above files.  If
"Protobuf" provides a separate development package or SDK, be sure it has
been installed.
````

https://github.com/systemed/tilemaker/actions/runs/6598962033/job/17927614211

systemed:
This is now fixed and the Docker build is working again.

PR #565, opened 2023-11-03
Update Mac install instructions

systemed:
- Suggest cmake for Mac build
- Suggest Lua 5.3 and update Makefile to find it in Homebrew on ARM Macs


systemed:
#636 improves Lua detection in the Makefile, and #623/#625 remove the protobuf dependency, so this is obsolete.

PR #567, opened 2023-11-04
Fix Docker dynamic build

Booligoosh:
Fixes the dynamically linked Docker image by installing the required libs on the final Docker image as well as the build Docker image. Now that it's not statically linked, they need to be present on the final Docker image as well, so that the Tilemaker executable can access them.

Note that building it like this with dynamic linking and the dependencies installed on the image increases the size of the image to 444MB, undoing some of the work from #363.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04 AS src
+FROM debian:bullseye-slim AS src
 LABEL Description="Tilemaker" Version="1.4.0"
 
 ARG DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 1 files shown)


Booligoosh:
Note that the "Generate mbtiles" check is still failing, because the GitHub Action is using the `master` image from Docker.

This probably ties into a larger conversation around tagging, versioning, and release management. Ideally, the CI would only run when new version tags are pushed, and then upload the Docker image with the correct version tag.

Then the GitHub action can reference the docker image tagged with the same version as the action itself - so if someone's workflow file references  `systemed/tilemaker@v2.0.0` then it'll use the image `docker://ghcr.io/systemed/tilemaker:v2.0.0` rather than `docker://ghcr.io/systemed/tilemaker:master`.

systemed:
Superb - thank you. I've merged it and forced a re-run (so it picks up the new image) and it's happily working.

Yes, that sounds like a good direction for the future for the image management.

PR #568, opened 2023-11-05
free SHPObject when done with it

cldellow:
It looks like Tilemaker converts shapefile objects into its own representation, after which point I think we can safely free the shape objects. I don't think `SHPClose` is sufficient to free the individual objects.

For the shapefiles used by the Shortbread tile set (1.4GB of coastline .shp files), this seems to decrease peak memory usage by ~1.7G


systemed:
Very good spot - thank you!

PR #572, opened 2023-11-05
clear UsedWays sooner

cldellow:
This is only a small improvement.

Previously, the memory for UsedWays is freed after all phases. This PR changes it to be freed before the final phase. For non-renumbered PBFs, UsedWays takes up 256MB, so giving that back is somewhat meaningful.

It may not make a practical difference in being able to process a larger PBF--I haven't dug too much into the memory use of the `Relations` phase to know if it uses much incremental memory.

This PR also removes the `ReadPhase::All` phase, which seems to be unused. I'm not sure if it's still in the code as a useful debugging thing, so please correct me if removing it was bad.

### DIFF
--- a/include/osm_store.h
+++ b/include/osm_store.h
@@ -505,6 +505,7 @@ public:
 	void ensure_used_ways_inited() {
 		if (!used_ways.inited) used_ways.reserve(use_compact_nodes, nodes_size());
 	}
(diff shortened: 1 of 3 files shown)


cldellow:
This seems like relatively small beer, given that it's a fixed cost, and if you really care about memory, you'll be using `--compact`, so closing.

PR #577, opened 2023-11-11
Avoid copying strings/tag maps

cldellow:
`OSMLuaProcessing`'s `Find`, `setWay`, `setNode` and `setRelation` functions are called a lot, so avoiding copies is helpful. Avoiding these copies results in a savings of ~1.5% of the time from startup to ready-to-write-output for my usage.

I believe the changes are safe: the tag map is only accessed for the duration of `setWay`, `setNode` and `setRelation`, so the pointer to the map in the caller's scope is fine.

I believe returning a reference from `::Find` is also OK: for `std::string`, the relevant code in Kaguya is: https://github.com/satoren/kaguya/blob/9d77cad73846f5d8481f3f7e86e8c53c4891c2fb/include/kaguya/type.hpp#L654-L657, which uses `lua_pushlstring` to pass it to Lua code.

And according to https://www.lua.org/pil/24.2.1.html:

> Lua never keeps pointers to external strings (or to any other object, except to C functions, which are always static). For any string that it has to keep, Lua either makes an internal copy or reuses one. Therefore, you can free or modify your buffer as soon as these functions return.

### DIFF
--- a/include/osm_lua_processing.h
+++ b/include/osm_lua_processing.h
@@ -92,7 +92,7 @@ public:
 	bool Holds(const std::string& key) const;
 
 	// Get an OSM tag for a given key (or return empty string if none)
-	std::string Find(const std::string& key) const;
+	const std::string& Find(const std::string& key) const;
 
 	// ----	Spatial queries called from Lua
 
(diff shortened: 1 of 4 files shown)


systemed:
Another good one and seems to work flawlessly - thanks!

PR #578, opened 2023-11-12
Re-use ifstream and OSMLuaProcessing

cldellow:
Previously, these were generated four times per block.

Now they're thread-locals, so they're generated once per phase, independent of the number of blocks.

This seems to be important for PBFs from Geofabrik, which have many, small blocks vs BBBike:

- Geofabrik GB extract: 26,000 blocks, 65KB/block
- BBBike partial USA extract: 33 blocks, 16,000KB/block

When processing Geofabrik extracts, I noticed that my CPU was idler than I'd expect during the PBF reading phase. strace suggested that Lua initialization takes an expensive lock, blocking other readers who are also trying to initialize Lua.

PBF reading for Great Britain is a bit faster with this change:

Before:

real 1m48.685s
user 21m10.017s
sys 1m36.528s

After:

real 1m36.549s
user 19m52.995s
sys 0m54.015s

### DIFF
--- a/include/read_pbf.h
+++ b/include/read_pbf.h
@@ -26,8 +26,8 @@ public:
 
 	PbfReader(OSMStore &osmStore);
 
-	using pbfreader_generate_output = std::function< std::unique_ptr<OsmLuaProcessing> () >;
-	using pbfreader_generate_stream = std::function< std::unique_ptr<std::istream> () >;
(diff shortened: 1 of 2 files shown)


systemed:
That does seem to speed up pbf reading noticeably - excellent, thank you.

PR #583, opened 2023-11-17
AttributeStore memory tweaks

cldellow:
This fixes a bug where:

```lua
obj:Attribute('name', 'some value')
obj:Attribute('name', 'some other value')
```

sometimes resulted in the first value being saved, sometimes the second. Now the last value always wins.

There are also some memory reductions: about 2,300MB less memory is needed for GB. This comes from two locations:

- AttributeStore - 1,300MB
- OsmMemTiles - 1,000MB

The PR does a few things.

1. Rather than store a whole `std::string` in `AttributePair` to identify the key (`name:latin`, `kind`, `iata`, etc), it identifies the key by a `uint16_t`. This limits Tilemaker to only 64K keys -- probably fine? Shortbread uses about 50, for reference.
2. Rather than store a whole `AttributePair` in an `AttributeSet`, it references them by a vector of `uint32_t`. This limits Tilemaker to 4B `AttributePair`s. This might be an issue -- but it's easily extended to more if needed. I see elsewhere that [Tilemaker is limited to 1B `AttributeSet`s](https://github.com/systemed/tilemaker/blob/d9bb72b929226937c9b06bd5279b0ec09a162a53/include/output_object.h#L42) -- probably if one has an issue, the other will, too.
3. I lied, `AttributeSet` only sometimes uses an actual vector of `uint32_t`s -- most of the time, a vector is overkill, as items only have a handful of attributes. Instead, `AttributeSet` typically uses the 24 bytes that a vector would occupy to store an array of 4 shorts and 4 ints†. Only if this array gets exhausted does it change to using a vector.
4. `vector_tile::Tile_Value` turns out to be quite big -- it's 96 bytes. I replaced it with a union of `std::string`, `float` and `bool`. I didn't notice any negative impact to PBF generation at the end from newing it up on a per-layer basis.
5. Recognize when a user calls `Layer` with the same geometry, and avoid processing/storing the geometry a second time. For example, when a profile writes a geometry for a river as well as for its name.

† The system tries to assign IDs that would fit in a short to those `AttributePair`s that it thinks will be very popular, like `rank=1`, `amenity=toilets`, `indoor=false` and so on. This lets most `AttributeSet`s store their pairs without having to allocate a vector.

### DIFF
--- a/include/attribute_store.h
+++ b/include/attribute_store.h
@@ -411,7 +411,7 @@ struct AttributeStore {
 	int lookups=0;
 
 	AttributeIndex add(AttributeSet &attributes);
-	std::set<AttributePair, AttributePairStore::key_value_less> get(AttributeIndex index) const;
+	std::vector<AttributePair> get(AttributeIndex index) const;
 	void reportSize() const;
 	void doneReading();
 
--- a/include/attribute_store.h
+++ b/include/attribute_store.h
@@ -3,6 +3,7 @@
 #define _ATTRIBUTE_STORE_H
 
 #include <mutex>
+#include <deque>
 #include <iostream>
 #include <boost/functional/hash.hpp>
 #include <boost/container/flat_map.hpp>
--- a/include/attribute_store.h
+++ b/include/attribute_store.h
@@ -370,6 +370,7 @@ struct AttributeSet {
 	}
 
 	void addPair(uint32_t index);
+	void removePairWithKey(const AttributePairStore& pairStore, uint32_t keyIndex);
 private:
 	void setValueAtIndex(size_t index, uint32_t value) {
 		if (useVector) {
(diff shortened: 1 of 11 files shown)


systemed:
This looks really good - thank you. I'll have a proper look at it this weekend. I like the optimisation for popular pairs - very clever.

systemed:
Great Britain:
- master: 7:21.83, 14.5GB
- refactor_geometries: 7:00.72, 14GB
- this PR: 6:33.40, 11.5GB

Germany:
- master: 13:56.06, 30.1GB
- master with --store: 15:06.89, 28.1GB
- this PR: 12:29.84, 23.7GB
- this PR with --store: 12:54.88, 21.6GB

That's an amazing improvement. Thank you!

PR #585, opened 2023-11-17
Document new optional argument minzoom of node:Attribute and way:Attribute function

Nakaner:
The feature was introduced by #219.


systemed:
Thank you!

PR #586, opened 2023-11-18
subdivide blocks if needed in Phase::Relations

cldellow:
This is a small quality-of-life tweak for people iterating on a theme on a small extract.

I often test on a Nova Scotia extract. It's big enough to test many things, but small enough to be fast to process, about 25 seconds. 6 seconds are spent in Phase::Relations processing.

The extract has 1,647 blocks, split roughly:
- 1,500-ish with nodes
- 140-ish with ways
- only 3 with relations

Thus, most of my CPU cores are idle during Phase::Relations, as there are not enough blocks to go around.

This is doubly bad, as a block of relations takes much longer than a block of nodes or ways, due to the more complex geometries involved.

This PR changes Tilemaker such that if it sees the relations phase has too few blocks, it artificially subdivides them in order to keep the CPUs busy.

This reduces my time for Nova Scotia to about 20 seconds.

### DIFF
--- a/include/read_pbf.h
+++ b/include/read_pbf.h
@@ -15,7 +15,7 @@
 class OsmLuaProcessing;
 
 struct BlockData {
-	uint64_t offset;
+	long int offset;
 	google::protobuf::int32 length;
 	// For Phase::Relations -- only process the entries in the block if
 	// index % denominator == modulo. This allows us to use multiple cores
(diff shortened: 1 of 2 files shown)


cldellow:
This will have conflicts with https://github.com/systemed/tilemaker/pull/590, so I revived it over there in https://github.com/systemed/tilemaker/pull/590/commits/1f38c9af75b6fe76aa58bae7e094b6101491a76c.

PR #591, opened 2023-11-21
tile index: don't block on shared lock

cldellow:
I promise I'm running out of PRs. :)

Previously, each worker thread tried to eagerly insert their object into the index, blocking if another thread had the lock.

The index update operation can be expensive -- it potentially has to rebalance the `map`, and potentially has to grow + copy the value's `vector`.

But I think no one reads the index until the output stage, so the workers could just add their updates to a queue and move on with their other work.

That's what they now do. If, after adding the object, the queue is looking large, one worker will self-nominate to flush the queue into the map. Two queues are maintained, so that other workers aren't blocked while this flush is happening.

`refactor_geometries` branch for reading GB:

```
real 1m28.080s
user 19m26.234s
sys 0m28.319s
```

This branch:

```
real 1m20.228s
user 18m59.398s
sys 0m16.494s
```

### DIFF
--- a/include/osm_lua_processing.h
+++ b/include/osm_lua_processing.h
@@ -193,6 +193,10 @@ public:
 
 	struct luaProcessingException :std::exception {};
 
+	void FlushTileIndex() {
+		osmMemTiles.FlushTileIndex();
+	}
+
 private:
 	/// Internal: clear current cached state
 	inline void reset() {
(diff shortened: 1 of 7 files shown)


cldellow:
Per https://github.com/systemed/tilemaker/discussions/593#discussioncomment-7667528, the global `TileIndex` should just cease to exist. Will make a new PR that does that instead.

PR #599, opened 2023-12-03
AttributeStore: skip locks when outputting

cldellow:
The locks in AttributeStore are necessary only during PBF reading, to avoid concurrent mutations corrupting things.

Once we're writing the mbtiles, it's safe to read without acquiring the lock. This eliminates ~9% of system time, and ~2-3% of wall clock time.

The PR also adds a `finalize()` to AttributeStore, AttributeKeyStore and AttributePairStore. Nothing actually uses this yet - I initially checked the `finalized` variable and threw if an unsafe method was called, but that gave up the speed benefits, so I removed it again.

Perhaps in the future, a debug build could leave such checks in to detect programming errors.

### DIFF
--- a/include/attribute_store.h
+++ b/include/attribute_store.h
@@ -23,12 +23,15 @@ struct string_ptr_less_than {
 
 class AttributeKeyStore {
 public:
-	AttributeKeyStore(): keys2indexSize(0) {}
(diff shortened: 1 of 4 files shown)


systemed:
Looks good - no issues that I can see. Thanks!

PR #606, opened 2023-12-05
Clipping monster polygons

cldellow:
This implements some ideas to mitigate the impact of clipping monster multipolygons.

In particular, for any multipolygon that is fully clipped out (i.e. we're in its hole) or clips to the full extent of a tile at z6..z8, we'll track the z9 tiles that it would fully cover or be excluded from.

Later, when processing a tile at z7 or higher, we'll check this tracking information to short-circuit doing an expensive clip operation.

Processing the Hudson Bay extract [1]:

master:
```
real	22m27.753s
user	336m50.481s
sys	0m25.314s
```

this branch:
```
real	3m46.838s
user	45m33.959s
sys	0m12.161s
```

I think the general idea here is sound...but I also think I don't know much about geometry or how it interacts with vector tiles, so a critical eye & mind would be extra appreciated.

Eyeballing the resulting tiles, it seems to look OK. This is not very rigorous, though. :) I wonder if there is tooling that will diff an mbtiles or pmtiles file? In theory, the files ought to be equivalent.

[1]: https://extract.bbbike.org/?lang=en;sw_lng=-99.269;sw_lat=53.617;ne_lng=-65.244;ne_lat=64.616;format=osm.pbf;city=a%20big%20bay;email=cldellow%40gmail.com;as=76.54103151056417;pg=0.8698404751901266;coords=;oi=1;layers=B000T;submit=;expire=

### DIFF
--- a/include/tile_data.h
+++ b/include/tile_data.h
@@ -248,7 +248,7 @@ protected:
 	uint16_t z6OffsetDivisor;
 
 	// Guards objects, objectsWithIds.
-	std::vector<std::mutex> objectsMutex;
+	mutable std::vector<std::mutex> objectsMutex;
 	// The top-level vector has 1 entry per z6 tile, indexed by x*64 + y
 	// The inner vector contains the output objects that are contained in that z6 tile
 	//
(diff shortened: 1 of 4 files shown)


cldellow:
>  (i.e., we're in its hole)

Ah, I suppose another cause of this (and a more likely one?) is that we're just outside of the polygon, not in one of its holes. I think this is possible because we use a bounding box when deciding which tiles the polygon affects.

I shouldn't be trusted with geometry stuff.

cldellow:
Thanks for the pointer to those tickets!

Yes, if you squint, this PR seems in the vein of what you mention in https://github.com/systemed/tilemaker/pull/479#issuecomment-1481408713:

> store a tile bitmap with each rtree (as a vector, though this of course will increase memory consumption)

The r-tree currently answers the question "does this object possibly output to a given tile?"

I guess this would widen the r-tree's capacities to "does this object definitely output to a given tile?" and, even further "when it does output to a given tile, does it cover it completely?"

Doing the pruning earlier is attractive. This also looks like it'd be in code near `fillCoveredTiles`, which I've been wondering about. I occasionally see it pop up in stack traces when my computer is only using 1 core at 100% -- Antarctica and the Hudson Bay extract both have this problem when reading the biggest of their relations.

I'd want to measure the memory impact -- this PR doesn't yet try to do something intelligent with memory, but it should be straight-forward to turn this PR's maps into a least-recently-used cache so that bitmaps from past z6 tiles get evicted naturally. The large object index, by contrast, is global and long-lived.

I've put this PR as draft while I investigate.

I also noticed two other things:
1. the PR has a bug - z9/145/160 incorrectly excludes the bay
2. some rough prototyping makes me think we can get another 70% decrease in writing if we extend this idea to the z10..z13 tiles, too

cldellow:
Closing in favour of https://github.com/systemed/tilemaker/pull/607

Issue #596, opened 2023-11-29
Low CPU usage when writing tiles


systemed:
When processing an Antarctica extract, CPU is sitting at around 120% on a 16-core machine during much of tile writing. 

This suggests we have a lock issue of some sort - either one of our own mutexs or sqlite. Antarctica is characterised by large areas of emptyish tiles so there’s less geometry processing going on than usual.

We could consider batching sqlite inserts if this proves to be the cause.

systemed:
#598 has made this significantly better. Rather than sitting at 120%, it's now closer to 280% or so. I'll leave this open in case any other improvements come to mind but I think we've probably plucked the low-hanging sqlite fruit.

PR #607, opened 2023-12-07
Clipping monster polygons 2

cldellow:
This replaces https://github.com/systemed/tilemaker/pull/606

When writing, we keep a cache of previously clipped geometries. Higher zooms will prefer to clip a cached, clipped geometry over an uncached, unclipped geometry.

With this PR:

- Hudson Bay: 1m28s
- Antarctica: 9m5s

Antarctica doesn't max out the CPU due to https://github.com/systemed/tilemaker/issues/596 - I think a next step there could be to have a buffer so workers can just enqueue a tile (up to some limit) and move on with their work.

I also noticed that the Antarctica mbtiles couldn't be converted by the `pmtiles convert` command. I suspect this is a pre-existing issue, but I've never generated Antarctica before. My guess is it's related to https://github.com/systemed/tilemaker/blob/06b68df2a969fc6ddc1e579ee6747458661d8683/src/tile_worker.cpp#L187, and we need to not write a tile at all if it has no features, not just remove the empty feature.

### DIFF
--- a/include/coordinates.h
+++ b/include/coordinates.h
@@ -33,6 +33,15 @@ public:
 			return false;
 		return y == obj.y;
 	}
+
+	bool operator <(const TileCoordinates_ & obj) const
+	{
+		if (x != obj.x)
+			return x < obj.x;
+
+		return y < obj.y;
+	}
+
 };
 struct TileCoordinatesCompare {
     bool operator()(const class TileCoordinates_& a, const class TileCoordinates_& b) const {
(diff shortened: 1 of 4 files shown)


cldellow:
This and https://github.com/systemed/tilemaker/pull/590 lets me generate North America in 43 minutes! The resulting mbtiles was able to be converted to a pmtiles file, which gives me some comfort that the Antarctica thing is probably unique to Antarctica.

Runtime looked like it was also impacted by #596, so hopefully we can do better than 43 minutes, too.

systemed:
This is a really neat, elegant solution! Thank you.

PR #608, opened 2023-12-07
buffer sqlite writes if needed

### LINKED ISSUE #596: Low CPU usage when writing tiles

cldellow:
Fixes #596, I think.

Time to process Antarctica drops from 9m5s to 3m51s

### DIFF
--- a/include/mbtiles.h
+++ b/include/mbtiles.h
@@ -7,6 +7,14 @@
 #include <vector>
 #include "sqlite_modern_cpp.h"
 
+struct PendingStatement {
+	int zoom;
+	int x;
+	int y;
+	std::string data;
+	bool isMerge;
+};
+
 /** \brief Write to MBTiles (sqlite) database
 *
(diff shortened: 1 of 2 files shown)


systemed:
This is great - CPU happily pegged at 16x100% throughout!

I'm getting a consistent "Command terminated by signal 11" right at the end of tile generation when processing Antarctica (including landcover). I put a bit of logging into `closeForWriting` but it's dying before that's called; I'll add a few more earlier in the process and let you know if I find anything. It doesn't happen with the smaller areas I've tried (Oxfordshire and Great Britain).

systemed:
Ah - that would make sense why it stopped after #607. I'd wondered about valgrind but the thought of it slowly crunching through the whole of Antarctica put me off!

PR #620, opened 2023-12-19
PMTiles support

systemed:
This adds support for generating [.pmtiles](https://github.com/protomaps/PMTiles) archives. Format is autodetected from the output filename (i.e. `.pmtiles`->PMTiles, `.mbtiles`->MBTiles).

This does not produce the most efficient .pmtiles archives possible. Threading means tilemaker's tile output order isn't sequential, so we can't set `.clustered`. We also don't do anything smart with optimising directory sizes.

Memory usage and execution time will be higher than with .mbtiles. In particular, we need to keep an index of the file location (offset) of each tile we write. This is then compiled into root/leaf directories after tile generation. There's probably scope to improve this!

Some implementation details:

- File order is header, root directory, tile data, leaf directories, metadata.
- We keep an index of any tiles under 100 bytes (uncompressed), such as a typical sea tile, so we only need to write these out once.
- Tiles below z6 are written to the root directory. All other tiles go to leaf directories with a maximum size of 10000000 entries.
- For areas of <100 z6 tiles (Europe and below), we use a map for the tile offset index. For larger areas (planet), we use a vector.
- Compression is always gzip.
- There's a small amount of refactoring to move MBTiles-specific code out of tilemaker.cpp.

[PMTiles Viewer](https://protomaps.github.io/PMTiles/) seems happy with the results at county level. I've successfully generated a Great Britain .pmtiles and verified it with `pmtiles verify` (both with sparse and dense index) but not viewed it yet.

/cc @bdon

### DIFF
--- a/include/pmtiles.h
+++ b/include/pmtiles.h
@@ -19,6 +19,8 @@ struct TileOffset {
 #define HEADER_ROOT 16384
 // Tile ID at which to start using leaf directories (=z6/0/0)
 #define FIRST_LEAF_TILE 1365
+// Threshold for using the root directory only
+#define ROOT_ONLY 2200
 // Maximum size in bytes of tiles considered "tiny" (i.e. potentially repeatable)
 #define TINY_LENGTH 100
 // Expire the tiny cache when it reaches this size
--- a/include/pmtiles.h
+++ b/include/pmtiles.h
@@ -2,6 +2,19 @@
 #ifndef _PMTILES_H
 #define _PMTILES_H
 
+#include <fstream>
+#include "external/pmtiles.hpp"
+
+struct TileOffset {
+	uint64_t offset : 40;
+	size_t length : 24;
+
+	TileOffset(uint64_t offset, size_t length) : offset(offset),length(length) { }
+	TileOffset();
+};
+
+#define LEAF_DIRECTORY_SIZE 10000000
+
 class PMTiles { 
 //	sqlite::database db;
 //	std::vector<sqlite::database_binder> preparedStatements;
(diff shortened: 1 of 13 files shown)


include/pmtiles.h
```diff
@@ -0,0 +1,57 @@
...
+
+	TileOffset(uint64_t offset, size_t length) : offset(offset),length(length) { }
+	TileOffset();
+};
+
+// Maximum number of tiles in a leaf directory
+#define LEAF_DIRECTORY_SIZE 10000000
+// Combined size of header and root directory (= start of tile data)
+#define HEADER_ROOT 16384
+// Tile ID at which to start using leaf directories (=z6/0/0)
```
bdon, line 20: hardcoding this might not be useful in the case where you have a deep tileset that still happens to fit completely within the root directory - ideal for city-sized extracts to remove the overhead of the additional fetch

include/pmtiles.h
```diff
@@ -0,0 +1,57 @@
...
+#define FIRST_LEAF_TILE 1365
+// Maximum size in bytes of tiles considered "tiny" (i.e. potentially repeatable)
+#define TINY_LENGTH 100
```
bdon, line 23: in planetiler our experiments determined the tile size in bytes isn't well correlated to repeatable tiles because of tag data - instead, detect a small # of square features

bdon:
left a comment inline, to summarize:

> Tiles below z6 are written to the root directory. All other tiles go to leaf directories with a maximum size of 10000000 entries.

You definitely want an all-root archive whenever possible for city-sized tilesets

> Threading means tilemaker's tile output order isn't sequential, so we can't set .clustered

The only consequences here should the directories take up more bytes, and you can't use `pmtiles extract` on an output. For cloud storage there isn't a huge locality advantage in accessing nearby parts of the same file

systemed:
> > Tiles below z6 are written to the root directory. All other tiles go to leaf directories with a maximum size of 10000000 entries.
> 
> You definitely want an all-root archive whenever possible for city-sized tilesets

Good call. Changed in https://github.com/systemed/tilemaker/pull/620/commits/03f857518603baa41333e242576781652b4550b9 - it will now use just the root directory below a certain number of entries (set to 2200 which seems to work reliably and still leave a bit of headroom).

> > Threading means tilemaker's tile output order isn't sequential, so we can't set .clustered
> 
> The only consequences here should the directories take up more bytes, and you can't use `pmtiles extract` on an output. For cloud storage there isn't a huge locality advantage in accessing nearby parts of the same file

Ah, that's good to know.

include/pmtiles.h
systemed: That's interesting. Should be doable by tweaking the various `tile_worker` code a bit but I think I'll leave that for a further enhancement.

PR #604, opened 2023-12-04
Lua interop

cldellow:
(This is meant to be merged after https://github.com/systemed/tilemaker/pull/623. Until then, you can see the lua-interop-specific diffs here: https://github.com/cldellow/tilemaker/compare/protozero...cldellow:lua-interop?expand=1)

A mix of tweaks to avoid Lua overhead - reading commit by commit is probably the easiest path to review.

PR #624, opened 2023-12-27
use vtzero instead of libprotobuf

cldellow:
This is meant to be merged after https://github.com/systemed/tilemaker/pull/604. Until then, you can see the vtzero-specific diffs here: https://github.com/cldellow/tilemaker/compare/lua-interop...cldellow:vtzero?expand=1 (I can rebase this against something other than lua-interop, if needed.)

Switch from libprotobuf to vtzero, and fix a few places where we copied the geometries during writing.

vs lua-interop branch, this branch is about 7% faster to process GB: 1m43s vs 1m51s

I don't use --merge, so my testing of the --merge mode was somewhat artificial. I built an mbtiles, then built it again with --merge and checked that features seemed duplicated.


cldellow:
Bah, git isn't very happy about merging master's squash commits. I'm going to re-open this against master with the relevant commits cherry-picked.

PR #626, opened 2023-12-28
lua interop

cldellow:
Replaces #604 

This PR reduces overhead when calling between Lua and C++:

- use the global namespace, e.g. `Find()` vs `way:Find()`
- use a specialized class for looking up tags
- don't use `string` as arguments when calling `Find()` or `Holds()`

Also some non-Lua related improvements to reduce lock contention on machines with many cores


cldellow:
Merged into v3 branch: e768595c4b18a7bf6dfd441a35510a3e6cf8a9b0

Issue #402, opened 2022-03-23
node_keys should be optional


systemed:
Currently tilemaker will error if `node_keys` isn't supplied in process.lua, even if you're just processing shapefiles and no .pbf.

systemed:
Fixed in v3.

Issue #337, opened 2021-10-06
Bus and metro stations not showing up?


wimpie3:
I'm not able to show public transport stations on my map. POI's like hospitals, parkings and shops appear, but no stations or bus stops.

My layout json uses the "poi" source layer with a filter on "bus", "rail","airport".

Is Tilemaker using an alternative name for the transportation layer? Or does it currently skip those points while making the vector layers?

systemed:
Ah, I see. Try building #338 and see if that makes it better.

wimpie3:
I'm on Windows so that will be difficult...

systemed:
You don't need to build tilemaker itself, it's just a change to the .lua file. If you change the process-openmaptiles.lua file to have that fix, then rerun tilemaker, it should (fingers crossed) work for you.

wimpie3:
Seems to be working! Thank you!

systemed:
Excellent!

systemed:
Can you give me an OSM object link, please. For example, https://www.openstreetmap.org/node/123456

wimpie3:
That is bizarre, cause the tram icons I can see on the maputnik link aren't visible on the openstreet map! OSM only shows bus stops...

wimpie3:
So this should work, right? Hmmm... it's not...

```
      "id": "poi-tram_stop",
      "type": "symbol",
      "source": "openmaptiles",
      "source-layer": "poi",
      "minzoom": 15,
      "filter": [
        "all",
        [
          "==",
          "$type",
          "Point"
        ],
        [
          "==",
          "class",
          "railway"
        ],
        [
          "==",
          "subclass",
          "tram_stop"
        ]
      ],
```

systemed:
Don't know, I'm afraid - debugging styles is a bit outside tilemaker's scope ;)

I'd recommend using `vt2geojson` as above for debugging - you can capture a vector tile in GeoJSON format and then look through it with a text editor to see if the data you want is in there.

wimpie3:
tilemaker --input belgium-latest.osm.pbf --output=tiles/ --process process-openmaptiles.lua --config config-openmaptiles.json 

Source of the PBF is from geofabrik.

Extract ends with "Sorting  nodes" and returns to the command prompt. I've retried it with the original lua script and it's failing as well. I've recently installed Windows 11 on my laptop, might be due to that?

Issue #190, opened 2020-12-06
Filter out ways when importing


ttsirkia:
If I have understood correctly, while reading the pbf file `tilemaker` will store all the ways to the memory. Based on the number of ways to be found, the memory consumption can be high.

What if I needed only, let's say motorways or land usage information, would it be possible to have some kind of filter Lua function to skip those ways which are not relevant? Does it already work like that for nodes with the `node_keys` list?

I think this would reduce the memory consumption quite a lot if the whole information is not needed but are there some drawbacks or something that I didn't understood correctly about the import process?

ttsirkia:
OK. I made some experiments with Estonian map data as it is not a tiny but not too big.

With the shipped OSM process files:
`Stored 10045947 nodes, 1165834 ways, 3396 relations`

With the shipped almost blank process files:
`Stored 10045947 nodes, 1092984 ways, 303 relations`

With an empty process file which does not include anything:
`Stored 10045947 nodes, 84422 ways, 0 relations`

So the node number is always the same but ways seem to reduce. However, it is not zero in the last test. The `.osm.pbf` file is 83 MB, the memory consumption in the first two tests around 1 GB and in the last around 500 MB. Is there actually anything that could be reduced in the importing phase to drop out unnecessary data prior writing the `mbtiles` file?

systemed:
Yes, tilemaker reads all nodes in the .pbf first, then (multipolygon) relations, then ways. It needs to read nodes before ways so that, when processing a way, it has the location of the way available for functions like `way:Intersects` (and I'd like to add more functions like this, as per #167).

`node_keys` is just a speed optimisation that avoids calling the Lua code for uninteresting nodes - it doesn't affect memory usage.

I suspect the reason that the OMT and example process files have similar results is because both of them include buildings, which are the biggest contributor to OSM data bloat.

You can eliminate unwanted nodes from the .pbf in advance by using tools such as osmfilter/osmconvert or osmium. You can also potentially reduce memory requirements by preprocessing the .pbf using [mapsplit](https://github.com/simonpoole/mapsplit).

_Conceivably_ we could implement a `way_keys` function that did an extra first pass to read ways, and if they match the tags, note their waynodes. Only then would we then read the node lat/lon on the next pass. That's not something I have any plan to implement (osmfilter/osmconvert does the same and works for me) but I wouldn't reject a PR implementing it.

[...]

PR #629, opened 2023-12-29
generalize node_keys; add way_keys

cldellow:
This PR generalizes the idea of `node_keys`, adds `way_keys`, and fixes https://github.com/systemed/tilemaker/issues/402.

I'm not too sure if this is generally useful - it's useful for one of my
use cases, and I see someone asking about it in https://github.com/systemed/tilemaker/issues/190
and, elsewhere, in https://github.com/onthegomap/planetiler/issues/99

If you feel it complicates the maintainer story too much, please reject.

The goal is to reduce memory usage for users doing thematic extracts by
not indexing nodes that are only used by uninteresting ways.

For example, North America has ~1.8B nodes, needing 9.7GB of RAM for its node
store. By contrast, if your interest is only to build a railway map, you
require only ~8M nodes, needing 70MB of RAM. Or, to build a map of
national/provincial parks, 12M nodes and ~120MB of RAM.

Currently, a user can achieve this by pre-filtering their PBF using
osmium-tool. If you know exactly what you want, this is a good
long-term solution. But if you're me, flailing about in the OSM data
model, it's convenient to be able to tweak something in the Lua script
and observe the results without having to re-filter the PBF and update
your tilemaker command to use the new PBF.

Sample use cases:

```lua
-- Building a map without building polygons, ~ excludes ways whose
-- only tags are matched by the filter.
way_keys = {"~building"}
```

```lua
-- Building a railway map
way_keys = {"railway"}
```

```lua
-- Building a map of major roads
way_keys = {"highway=motorway", "highway=trunk", "highway=primary", "highway=secondary"}`
```

Nodes used in ways which are used in relations (as identified by
`relation_scan_function`) will always be indexed, regardless of
`node_keys` and `way_keys` settings that might exclude them.

A concrete example, given a Lua script like:

```lua
function way_function()
  if Find("railway") ~= "" then
    Layer("lines", false)
  end
end
```

it takes 13GB of RAM and 100 seconds to process North America.

If you add:

```lua
way_keys = {"railway"}
```

It takes ~2.5GB of RAM and 47 seconds.

Notes:

1. This is based on `lua-interop-3`, as it interacts with files that are
   changed by that. I can rebase against master after lua-interop-3 is
   merged.

2. The names `node_keys` and `way_keys` are perhaps out of date, as they
   can now express conditions on the values of tags in addition to their
   keys. Leaving them as-is is nice, as it's not a breaking change.
   But if breaking changes are OK, maybe these should be
   `node_filters` and `way_filters` or similar?

3. Maybe the value for `node_keys` in the OMT profile should be
   expressed in terms of a negation, e.g. `node_keys = {"~created_by"}`?
   This would avoid issues like https://github.com/systemed/tilemaker/issues/337 -- probably not very critical.

4. This also adds a SIGUSR1 handler during OSM processing, which prints
   the ID of the object currently being processed. This is helpful for
   tracking down slow geometries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,6 +94,7 @@ file(GLOB tilemaker_src_files
 	src/sharded_way_store.cpp
 	src/shared_data.cpp
 	src/shp_mem_tiles.cpp
+	src/significant_tags.cpp
 	src/sorted_node_store.cpp
 	src/sorted_way_store.cpp
 	src/tag_map.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,6 +96,7 @@ file(GLOB tilemaker_src_files
 	src/shp_mem_tiles.cpp
 	src/sorted_node_store.cpp
 	src/sorted_way_store.cpp
+	src/tag_map.cpp
 	src/tile_data.cpp
 	src/tilemaker.cpp
 	src/tile_worker.cpp
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,7 @@ tilemaker: \
 	src/sharded_way_store.o \
 	src/shared_data.o \
 	src/shp_mem_tiles.o \
+	src/significant_tags.o \
 	src/sorted_node_store.o \
 	src/sorted_way_store.o \
 	src/tag_map.o \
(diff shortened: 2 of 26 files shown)


systemed:
Absolutely happy with this - it's a good power-user feature and a handy optimisation for those using custom profiles. Thanks!

> But if breaking changes are OK, maybe these should be `node_filters` and `way_filters` or similar?

Given that we're breaking the Lua interface anyway then yes, let's go for `_filters`.

cldellow:
Merged into v3 branch: 9080cdb10e6afe98f3b3e36391dd2ca356d0af4e

PR #630, opened 2023-12-30
GeoJSON as alternative to shapefiles

systemed:
This enables GeoJSON files to be used in exactly the same way that shapefiles are.

Also refactor shapefile processing into its own class.


cldellow:
This and https://github.com/systemed/tilemaker/commit/b2785690d2d3cdc56fbe1835349c837316d60151 are very cool together. It came just in time to solve a problem I had been deferring solving.

Problem: I'm building a map where national/state parks are treated specially -- different features are emphasized vs the broader map. I still want the broader map to have these features in order to orient the user, but in the broader map, the features might only appear at higher zooms, or are styled with less contrast.

Solution with this PR:

1. Have a one-off Lua profile that emits only national/state parks polygons

2. https://github.com/cldellow/tilemaker/commit/0a6187719bc0e2c3aeb243d8af8b915fd31fa09a extends this branch so you can do `--output parks.geojson`. When writing, it will dump any materialized OSM multipolygons to GeoJSON.

3. Have a "real" Lua profile that consumes `parks.geojson` so that it can do `Intersects(...)` queries, and either adjust the output zooms, or emit an attribute that a future style rule can target.

My real profile does have this hacky bit in its config file:

```json
{
    "layers": {
        "parks": {
            "minzoom": 15,
            "maxzoom": 15,
            "source": "../parks.geojson",
            "index": true
        },
        ...
    },
    "settings": {
        "basezoom": 14
        ...
```

Since I'm only using the parks layer for spatial queries, I set its minzoom to a zoom greater than my map's basezoom to prevent it from being materialized into the output.

This is a bit of an OSM ouroboros...but it seems to work? Maybe it'll fall over if I scale it up.

(Maybe there is a better way to do this? I have close to zero familiarity with the GIS toolchain outside of OSM-related tools.)

cldellow:
Ah, related: when shp/geojson files are configured to be indexed, it looks like StoreGeometry needs a lock to avoid corrupting indexedGeometries/indices[layerName]/indexedGeometryNames

systemed:
Good spot - I’ve got a couple of tweaks including that which I’ll push tomorrow.

systemed:
Small push which fixes threading and another couple of issues.

> (Maybe there is a better way to do this? I have close to zero familiarity with the GIS toolchain outside of OSM-related tools.)

That's quite ingenious! Not particularly a priority but we could perhaps integrate that as a separate `--dump` option (rather than overloading `--output` which is intended for tile output).

The most common way to get individual features out of OSM in GeoJSON format is by using [Overpass Turbo](https://overpass-turbo.eu). But I'd guess you could also do it using `osmium extract` and then `ogr2ogr`.

cldellow:
Thanks for the pointers to Overpass and ogr2ogr.

Yes, this is fairly removed from Tilemaker's core reason for being. I'll let it kick around in my repo while I get a feel for it.

https://github.com/systemed/tilemaker/discussions/556 is another user who might be served by it (or something similar to it)

PR #631, opened 2023-12-30
C++ tileserver

systemed:
This is #301 brought up-to-date and with a couple of extra command-line options. It should help alleviate the pain-point people have if they don't already have Ruby up and running on their system.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,6 +123,6 @@ install(FILES docs/man/tilemaker.1 DESTINATION share/man/man1)
 
 add_executable(tilemaker-server server/server.cpp)
 target_include_directories(tilemaker-server PRIVATE include)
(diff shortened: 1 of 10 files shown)


pnorman:
Another option is Martin (including via docker), so we'd be looking at users who don't have ruby running or docker running.

systemed:
Indeed, and there are PHP equivalents and so on. The purpose of this is a “don’t make me think” way of quickly testing out your tilemaker-produced vector tiles, without having to identify and install an additional solution. For production I am sure people will continue to use one of the more fully featured servers available.

systemed:
Thanks again @bdon and sorry it's taken so long to merge this!

bdon:
Thanks for the update! Good to have more serving options... :P

PR #634, opened 2024-01-06
compile streamvbyte as C, not C++

cldellow:
(This fixes the ARM Mac build complaints about streamvbyte, one of the things reported at #536.)

In #590, I compiled these as C++ files.

This was wrong, as they're meant to be compiled as C files. Depending on your compiler, you would hit errors to do with `restrict` or `__restrict__` references.

This reverts the streamvbyte code to be a more vanilla copy of upstream, and updates the `Makefile` and `CMakeLists.txt` files to build them as C.

I've tested that this builds on ARM Ubuntu 22 with gcc, and @FunMiles has verified it builds on ARM Mac.


systemed:
Excellent - thank you! (Fine on Intel Mac too, unsurprisingly.)

PR #636, opened 2024-01-07
Smarter Lua detection in Makefile

### LINKED ISSUE #633: Improve Lua detection in Makefile
Ideally we should have something smarter than simply checking a list of likely paths.

Some prior art:

- https://github.com/zhaozg/lua-openssl/blob/master/Makefile
- https://github.com/moai/luamongo/blob/master/Makefile

systemed:
Detecting Lua is _hard_ and currently our Makefile takes a fairly blunt approach of just checking some likely paths.

This PR makes it a little smarter by trying to find the Lua executable and using that to detect version/implementation (i.e. Lua or LuaJIT). We can then use this knowledge to inform the paths we look in.

(CMake of course does all this automatically. I could have pulled in pkg-config but would rather not add the extra dependency.)

Tested on Ubuntu 20.04 and Intel MacOS 10.14 with a couple of different Lua setups.

Fixes #633.


systemed:
Merged into v3.

PR #641, opened 2024-01-12
remove PossiblyKnownTagValue; fixup LayerAsCentroid; fixup way_keys interaction with LayerAsCentroid; fixup Intersects

cldellow:
When I replaced #604 with #626, I botched extracting this part of the code. I had the trait, which taught kaguya how to serialize `PossiblyKnownTagValue`, but I missed updating the parameter type of `Attribute` to actually use it, so it was a no-op.

This PR restores the behaviour of avoiding string copies, but now that we have protozero's data_view class, we can use that rather than our own weirdo struct.

It also fixes an unrelated test build error in the way store test.

...and two issues to do with LayerAsCentroid's support for relation members.

...and an issue with the Intersects PR.

### DIFF
--- a/include/attribute_store.h
+++ b/include/attribute_store.h
@@ -10,6 +10,7 @@
 #include <boost/functional/hash.hpp>
 #include <boost/container/flat_map.hpp>
 #include <vector>
+#include <protozero/data_view.hpp>
 #include "pooled_string.h"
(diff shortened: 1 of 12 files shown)


systemed:
Thank you! Merged into v3.

PR #660, opened 2024-01-27
ignore --shard-stores if --compact is present

### LINKED ISSUE #657: "`Could not find node with id`" despite node presence in PBF
When running [tilemaker v3.0.0](https://github.com/systemed/tilemaker/releases/tag/v3.0.0) (release binary on Ubuntu 22.04) with the planet on debug profile / config, I get the following error:

```
tilemaker --config config.json --process process.lua --output planet.mbtiles --store planet.tmp --compact planet-osmium-catted.osm.pbf
[...]
terminate called after throwing an instance of 'std::out_of_range'                                                                                                                         
  what():  Could not find node with id 11498743532                                                                                                                                         
Command terminated by signal 6                                                                                                                                                             
```
`osmium getid`  confirmed that my input file does contain NodeID 11498743532:

```
<?xml version='1.0' encoding='UTF-8'?>
<osm version="0.6" generator="osmium/1.14.0">
  <bounds minlat="-90" minlon="-180" maxlat="90" maxlon="180"/>
  <node id="11498743532" version="1" timestamp="2024-01-08T00:58:56Z" uid="18863578" user="mountaineagle" changeset="146011008" lat="48.711566" lon="13.6249443"/>
</osm>
```

If I drop `--compact`, it runs through without a problem. This makes me think, it might have something to do with https://github.com/systemed/tilemaker/blob/3af62f6b33d6cf0372f215b6728003824c1ac027/src/node_stores.cpp#L93-L97.

cldellow:
Fixes #657 

These two settings are incompatible, because CompactNodeStore does not implement `.contains(...)`

Instead, if `--compact` is present, don't shard stores. Further, throw if code calls `.contains(...)`, so that we'll fail noisily and more clearly.

I debated teaching CompactNodeStore how to do `.contains(...)`, e.g. by maintaning a bitmap of which entries were present. Ultimately, though, I think sharding CompactNodeStores does not make sense -- each CompactNodeStore instance will use about the same amount of memory (vs the other stores, which will each use 1/N memory)

Also a small tweak: detect if `--compact` is used on a non-renumbered PBF, and warn the user.

### DIFF
--- a/include/node_stores.h
+++ b/include/node_stores.h
@@ -59,8 +59,8 @@ public:
 	void batchStart() override {}
 
 	// CompactNodeStore has no metadata to know whether or not it contains
-	// a node, so it's not suitable for used in sharded scenarios.
-	bool contains(size_t shard, NodeID id) const override { return true; }
+	// a node, so it's not suitable for use in sharded scenarios
+	bool contains(size_t shard, NodeID id) const override { throw std::runtime_error("CompactNodeStore does not support contains"); }
 	NodeStore& shard(size_t shard) override { return *this; }
 	const NodeStore& shard(size_t shard) const override { return *this; }
 	size_t shards() const override { return 1; }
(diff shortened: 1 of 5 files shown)


systemed:
Thanks! Yes, that makes sense.

Using `std::exchange` on a bool is a clever little idiom - never seen that before :)

PR #663, opened 2024-01-29
GeoJSON lines: skip leading whitespace in a chunk

cldellow:
D'oh, after running more GeoJSON lines content through my pipelines, I ran into a bug.

If the chunking algorithm returns a chunk that consists only of whitespace, the current code will try to parse at least one JSON object out of it, and fail.

This fixes that bug by consuming all the whitespace before trying to parse an item -- the `while(is.Tell() < chunk.length)` loop will no longer be entered, as there will be no non-whitespace content to consume.


systemed:
Thanks!

PR #668, opened 2024-02-04
GeoJSON writer: support points and linestrings

cldellow:
This fleshes out the todos in geojson_writer: Points, LineStrings and MultiLineStrings.


systemed:
Thank you! Laziness on my part not to have done them initially really... in my defence I was debugging some horrid multipolygons :)

cldellow:
Heh, no intention to call you out! I added these in my own fork while prototyping #669, and figured they'd be easy to extract into their own standalone PR

PR #672, opened 2024-02-07
AttributePair: use uint, not char

### LINKED ISSUE #671: Attribute(name, value, minzoom) writes attribute on all zoom levels
The minzoom argument of the function `Attribute(name, value, minzoom)` is not working.

How to reproduce:

* Download a motorway: `curl https://api.openstreetmap.org/api/0.6/way/4231749/full | osmium cat --input-format xml --output-format pbf -o w4231749.osm.pbf -`
* Create vector tiles from it: `./tilemaker --input w4231749.osm.pbf --output /tmp/tiles/ --bbox 8.2802,48.8782,8.3682,48.9414 --process resources/process-openmaptiles.lua --config resources/config-openmaptiles.json`
* Open the tile on zoom level 8 with `ogrinfo`: `ogrinfo -ro -al MVT:/tmp/tiles/8/133/88.pbf`

Ogrinfo will print a linestring geometry with a `surface` attribute. However, the Tilemaker process file says that the attribute is supposed to be written on zoom level 9 and larger only.

The bug was introduced by commit d62c480debf6c38cd5d2c6339bfaa995b05fb9f2 by @cldellow.

cldellow:
Fixes #671 

Since we only allocate 4 bits and `char` is signed, the usable space
was -8..7. Larger values (like, say, 12) overflow, and get interpreted
as a negative value, which means they don't act as a filter, since all
zoom values are natural numbers.

The tests didn't actually test that a zoom value could be roundtripped.
I updated them, and verified they failed before the code change, and
passed after the code change.

I've also allocated an extra bit so that we support minzooms up to z31,
vs just up to z15, since I think (?) some people generate up to z16

### DIFF
--- a/include/attribute_store.h
+++ b/include/attribute_store.h
@@ -42,13 +42,13 @@ private:
 	std::map<const std::string*, uint16_t, string_ptr_less_than> keys2index;
 };
 
-enum class AttributePairType: uint { Bool = 0, Float = 1, String = 2 };
+enum class AttributePairType: uint8_t { Bool = 0, Float = 1, String = 2 };
 // AttributePair is a key/value pair (with minzoom)
 #pragma pack(push, 1)
 struct AttributePair {
 	short keyIndex : 9;
 	AttributePairType valueType : 2;
-	uint minzoom : 5; // Support zooms from 0..31. In practice, we expect z16 to be the biggest minzoom.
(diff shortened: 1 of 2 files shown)


systemed:
Thanks for tracking that one down!

Issue #661, opened 2024-01-27
Latest version does not work on windows


Wicpar:
it just stops after one secons. 2.4.0 seems to work
```
PS \tilemaker-windows\build\RelWithDebInfo> ./tilemaker.exe --verbose --store ./store --config .\config-openmaptiles.json --process .\process-openmaptiles.lua --input planet-240115.osm.pbf --output tiles
Layer place (z0-14)
Layer boundary (z0-14)
Layer poi (z12-14)
Layer poi_detail (z14-14) -> poi
Layer housenumber (z14-14)
Layer waterway (z8-14)
Layer waterway_detail (z12-14) -> waterway
Layer transportation (z4-14)
Layer transportation_name (z8-14)
Layer transportation_name_mid (z12-14) -> transportation_name
Layer transportation_name_detail (z14-14) -> transportation_name
Layer building (z13-14)
Layer water (z6-14)
Layer ocean (z0-14) -> water
Layer water_name (z14-14)
Layer water_name_detail (z14-14) -> water_name
Layer aeroway (z11-14)
Layer aerodrome_label (z10-14)
Layer park (z11-14)
Layer landuse (z4-14)
Layer urban_areas (z4-8) -> landuse
Layer landcover (z0-14)
Layer ice_shelf (z0-9) -> landcover
Layer glacier (z2-9) -> landcover
Layer mountain_peak (z11-14)
Bounding box -180, -85.06, 180, 85.06
PS \tilemaker-windows\build\RelWithDebInfo>
```

Star-42:
Same here.
tilemaker just terminates after logging configured layers.

systemed:
I have just installed Windows on my Mac laptop and will see what I can work out. It won’t be quick as I’ll have to get VS Code, MSVC++ etc. up and running even before the debugging. But it’s on the radar.

qmjy:
same problem on v3.0.0

Star-42:
@systemd: Thanks for your quick response!
Meanwhile I manged to fix the issue.

source file: src/pbf_processor.cpp
method: PbfHasOptionalFeature
bool PbfHasOptionalFeature(const std::string& inputFile, const std::string& feature) {
	std::ifstream infile(inputFile, std::ifstream::in);
	auto header = reader.readHeaderFromFile(infile);
	infile.close();
	return header.optionalFeatures.find(feature) != header.optionalFeatures.end();
}

change:
  replace
	std::ifstream infile(inputFile, std::ifstream::in);
[...]

PR #677, opened 2024-02-10
SortedNodeStore::publishGroup: out of bounds read

cldellow:
Fixes the issued identified by @freeExec in #661

This code is a bit gross. It intentionally loops 1 past the size of the vector to ensure that we "publish" all the chunks.

There's a similar loop at lines 511-602.

In both loops, there's sort of three chunks of code, which I'll call "pre", "body" and "post". The pre/post sections need to be guarded, and the guard for the first loop's "post" section was missing.

Looking at it with fresh eyes, I think we could extract the "body" parts of both loops to their own functions, change the loop to use `<` instead of `<=` and then just call the body function again after the loop ends if it's needed. For now, I'm inclined to let a sleeping dog lie, but if more work happens in here, that'd be a good cleanup step.

This was also visible in valgrind, as it turns out--the error goes away with this change:

```
==1968107== Invalid read of size 4
==1968107==    at 0x33ABCD: SortedNodeStore::publishGroup(std::vector<std::pair<unsigned long, LatpLon>, std::allocator<std::pair<unsigned long, LatpLon> > > const&) (in /usr/local/bin/tilemaker)
==1968107==    by 0x33B715: SortedNodeStore::finalize(unsigned long) (in /usr/local/bin/tilemaker)
==1968107==    by 0x2FEE77: PbfProcessor::ReadPbfFile(unsigned int, bool, SignificantTags const&, SignificantTags const&, unsigned int, std::function<std::shared_ptr<std::istream> ()> const&, std::function<std::shared_ptr<OsmLuaProcessing> ()> const&, NodeStore const&, WayStore const&) (in /usr/local/bin/tilemaker)
==1968107==    by 0x140801: main (in /usr/local/bin/tilemaker)
```

SortedWayStore has a similar publishGroup function, but its internal implementation is different and it doesn't have this issue.

### DIFF
--- a/src/sorted_node_store.cpp
+++ b/src/sorted_node_store.cpp
@@ -423,9 +423,11 @@ void SortedNodeStore::publishGroup(const std::vector<element_t>& nodes) {
 			lastChunk = currentChunk;
 		}
 
-		tmpLatpLons[currentNodeIndex] = nodes[i].second.latp;
-		tmpLatpLons[currentNodeIndex + 256] = nodes[i].second.lon;
-		currentNodeIndex++;
+		if (i != nodes.size()) {
+			tmpLatpLons[currentNodeIndex] = nodes[i].second.latp;
+			tmpLatpLons[currentNodeIndex + 256] = nodes[i].second.lon;
+			currentNodeIndex++;
+		}
 	}
 
 	uint64_t chunks = currentChunkIndex;


cldellow:
I see that https://github.com/systemed/tilemaker/commit/c63040069824065a20f236524a0d53fbe9fccb27 failed with a segfault, looks like it was during node reading: https://github.com/systemed/tilemaker/actions/runs/7857409309/job/21441273093

It's possible that was this bug.

systemed:
Thanks!

> For now, I'm inclined to let a sleeping dog lie

🐶💤

PR #684, opened 2024-02-21
When serving tiles, add the CORS Access-Control-Allow-Origin: * header

amandasaurus:
This allows tiles served with `tileserver-server` to be easily used in simple web apps.


systemed:
Thank you!

PR #685, opened 2024-02-23
Stable tags for container images

leonardehrenfried:
For a while tilemaker has been building container images. This is great but so far only the very latest version is uploaded.

For my deployments I would not want to rely on a moving target but only reference a stable version.

This PR changes the container CI job so that it adds date-based tags to the container which will remain stable over time. They will have tags like `2024-02-23T12-42` and they will remain in the repo even if newer ones are added.

You can check what it looks like in this repo: https://github.com/leonardehrenfried/tilemaker/pkgs/container/tilemaker

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: "Continuous Integration"
 
 on:
   push:
-    branches: [ master ]
+    branches: [ stable-containers ]
   pull_request:
 
 env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -179,12 +179,12 @@ jobs:
 
       - name: Extract metadata (tags, labels) for Docker
         id: meta
-        uses: docker/metadata-action@5
+        uses: docker/metadata-action@v5
         with:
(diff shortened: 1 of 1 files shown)


daniel-j-h:
This is great work and we need this to make sure we can pin tilemaker to a specific version.

Have you thought about instead of date-stamping the docker images that we simply gitsha-stamp them? Then it would be relatively easy to go from a docker image tag to the repository and check out what's included. It would also allow us in the future to build staging versions, release candidates, and so on from pull requests.



And will we still have a latest default tag so that folks can get started without having to pick a specific tag?

leonardehrenfried:
In other projects I used to tag with both the SHA1 as well as the date. Then I realised that in several years of doing that I never needed to look up the SHA1 (not even once) and that's when I stopped adding it.

However, it's easy to do and I'm not against it.

The latest tag is already added by this PR.

leonardehrenfried:
Hello there, I'm giving this another go: @systemed would you be able to review this?

pnorman:
Another tag option is `git describe`'s output

PR #686, opened 2024-02-23
Reduce differences between official OpenMapTiles tiles and Tilemaker implementation

Nakaner:
This pull request reduces the differences between official OpenMapTiles tiles as served by MapTiler (the schema documentation lacks details, therefore I inspected their tiles) and the implementation of Tilemaker.

Changes:

* Add missing features classes to layer *landcover* (`natural=dune`, `natural=bare_rock/scree/scrub/shrubbery/tundra`).
* Add missing features to layer *place* (`place=province/borough/quarter/island/isolated_dwelling`).
* Add support for `place=island` mapped as polygons.
* Add attribute `subclass` for railway features, set correct attribute `class.
* Do not write railway features to layer *transportation_name*.
* Refactor layers *transportation_name*, *transportation_name_mid*, *transportation_name_detail*.
* Do not write roads without names or reference number to layer *transportation_name*.
* Do not write unnamed ferries to layer *transportation_name*.
* Set correct minzoom values for all roads in the *transportation* and *transportation_name* layer.
* Add `highway=raceway`.
* Fix handling of roads under construction.
* Do not set attribute `mtb_scale` if the tag is not set on the OSM way.
* Add aerialways to layer *transportation*.
* Do not set attribute `expressway` if the tag is missing.
* Add areas for pedestrians and platforms to layer *transportation*.
* Add `man_made=bridge` to layer *transportation*.

Some of the bugs were found by a client using a fork of the schema.

### DIFF
--- a/resources/process-openmaptiles.lua
+++ b/resources/process-openmaptiles.lua
@@ -227,10 +227,11 @@ landuseKeys     = Set { "school", "university", "kindergarten", "college", "libr
                         "retail", "stadium", "pitch", "playground", "theme_park", "bus_station", "zoo" }
 landcoverKeys   = { wood="wood", forest="wood",
                     wetland="wetland",
-                    beach="sand", sand="sand",
+                    beach="sand", sand="sand", dune="sand",
                     farmland="farmland", farm="farmland", orchard="farmland", vineyard="farmland", plant_nursery="farmland",
                     glacier="ice", ice_shelf="ice",
-                    grassland="grass", grass="grass", meadow="grass", allotments="grass", park="grass", village_green="grass", recreation_ground="grass", garden="grass", golf_course="grass" }
+                    bare_rock="rock", scree="rock",
+                    fell="grass", grassland="grass", grass="grass", heath="grass", meadow="grass", allotments="grass", park="grass", village_green="grass", recreation_ground="grass", scrub="grass", shrubbery="grass", tundra="grass", garden="grass", golf_course="grass", park="grass" }
 
 -- POI key/value pairs: based on https://github.com/openmaptiles/openmaptiles/blob/master/layers/poi/mapping.yaml
 poiTags         = { aerialway = Set { "station" },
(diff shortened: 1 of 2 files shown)


resources/process-openmaptiles.lua
```diff
@@ -405,117 +441,156 @@ function way_function()
...
-		Attribute("class", railway)
-		SetZOrder()
-		SetBrunnelAttributes()
-		if service~="" then
-			Attribute("service", service)
-			MinZoom(12)
-		else
-			MinZoom(9)
-		end
+		local class = railwayClasses[railway]
```
leonardehrenfried, line 552: I noticed that railway=proposed also ends up in the tiles. Am I right in thinking that this now takes care of it?

Nakaner: Neither the schema documentation nor the official implementation nor the tiles served by MapTiler contain `railway=proposed`.

`railwayClasses` contains all accepted values of `railway=*`.

leonardehrenfried: That was my point. It ended up in the tiles before this PR and presumably doesn't anymore, correct?

Nakaner: Yes, I will not end up in the tiles any more.

hyperknot:
I am curious about the licensing situation here. From what I see, OpenMapTiles development in the official repo has stopped (let's call it OMT 1.x) and now MapTiler is pushing into their private fork of it (let's call it OMT 2.x). 

In my understanding, this PR is essentially reverse engineering some changes from OMT 2.x and includes those back into OMT 1.x, but wired inside this repo.

But doing this might be entirely legal, depending on the license. Which is something I really don't know.

So the big questions is if MapTiler pulled a Mapbox here, or we are still allowed to somehow include the changes.

Nakaner:
@hyperknot I did not know which versions of OpenMapTiles exist at all. I found the sparse documentation that lacks many details. I had a client asking for these tiles. I produced the tiles with Tilemaker, they reported incompatibilities. I looked into the implementation and it answered some questions/confirmed some complaints. But the implementation was so difficult to read that I ended up looking in some vector tiles served by Maptiler (i.e. I reverse engineered it).

There are still a few differences left:

* the place ranking mentioned by @systemed
* usage of route relations for *transportation_name* layer
* unsure: disputed boundaries (I did not bother)

systemed:
Thank you!

PR #692, opened 2024-03-01
feat: export and render building:part

c84c:
First of all, thanks for this great project!

With this PR, I add the ability to export and render `building:part` relations (thanks to _Multipolygon relations are supported natively by tilemaker_).

For the 3D building render style, credits to @tgdbepe4 and his great [style](https://github.com/tgdbepe4/OSM-sac_scale-map-creating-with-tilemaker/blob/main/switzerland_style.json#L1422-L1440)!

The result is shown below for 2 monuments in Rome (Italy).

| Building | Before PR | After PR |
| :---         | :---:  | :---: |
| ["name:it": "Castel Sant'Angelo"](https://www.openstreetmap.org/#map=19/41.90289/12.46617) | CastelSantAngelo_before | CastelSantAngelo_after |
| ["name:it": "Piazza San Pietro"](https://www.openstreetmap.org/#map=18/41.90230/12.45733) | PiazzaSanPietro_before| PiazzaSanPietro_after|


pnorman:
OpenMapTiles says it [just uses building=*](https://openmaptiles.org/schema/#building), but later on goes to talk about the height of the building or building:part. I assume it's contradicting itself and in practice and the most common implementation includes building:part?

c84c:
Exploring [implementation](https://github.com/openmaptiles/openmaptiles/tree/master/layers/building) they use also `building:part` to insert buildings in mbtiles.

etienneJr:
Hi
Thanks for the work !

cartes.app users noticed an issue with this PR. It is [discussed here](https://forum.openstreetmap.fr/t/discussions-autour-des-fonds-carte-de-cartes-app/28339/45) in French.

It is coming from this recommandation [in the wiki](https://wiki.openstreetmap.org/wiki/Key:building:part#How_to_use):
> Where a building has been split into parts, a building outline has still to be drawn of the whole building way and should have the tags [building:levels](https://wiki.openstreetmap.org/wiki/Key:building:levels)=* and [height](https://wiki.openstreetmap.org/wiki/Key:height)=*. These values should be the maximum of all the building parts.

Therefore the building outline [way/90195611](https://www.openstreetmap.org/way/901956113) "hide" the building:part(s) because it is higher.

https://pmtiles.io/#url=https%3A%2F%2Fserveur.cartes.app%2Fpmtiles%2Fplanet.pmtiles&map=17.81/48.71687/2.191008
image

https://cartes.app/?allez=H%C3%B4pital+Paris+Saclay%7Cw1291813385%7C2.1915%7C48.7172#17.33/48.717586/2.190552/-59.2/75
image

The model in OSM is fine : https://beakerboy.github.io/OSMBuilding/?id=901956113
image

@c84c How can we modify the code to identify the building outline and set its height at 0 ? (we don't want to remove it since tags are on it)

Issue #664, opened 2024-01-30
MAC OS issue with types during the make command


amin212121:
I run make command on MAC OS but get the errors:

src/helpers.cpp:163:16: error: variable has incomplete type 'struct stat64'
        struct stat64 statBuf;
                      ^
src/helpers.cpp:163:9: note: forward declaration of 'stat64'
        struct stat64 statBuf;
               ^
src/helpers.cpp:164:6: error: no viable conversion from 'stat64' to 'int'
        int rc = stat64(filename.c_str(), &statBuf);
            ^    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.
make: *** [src/helpers.o] Error 1

<img width="1080" alt="image" src="https://github.com/systemed/tilemaker/assets/36712833/094d75ae-0317-42d2-b67e-957f188f2484">

systemed:
What version of macOS?

What processor (Intel or Apple Silicon)?

systemed:
...and I haven't tested it, but adding this near the top of src/helpers.cpp may work:

```c++
#if defined(__APPLE__)
#define stat64 stat
#endif
```

amin212121:
> What version of macOS?
> 
> What processor (Intel or Apple Silicon)?

Sonoma 14.1
Apple Silicon

amin212121:
> ...and I haven't tested it, but adding this near the top of src/helpers.cpp may work:
> 
> ```c++
> #if defined(__APPLE__)
> #define stat64 stat
> #endif
> ```

yes it helps. 
Thx

systemed:
That's great to know - thanks.

@cldellow Ok just to add that, do you think?

cldellow:
I think so!

There's probably another solution, too. It looks like perhaps this is only for the Makefile build, not the CMake build, which makes me think we must have subtly different include directory orders or something?

Sort of related: would you be open to a PR to add the Makefile builds to the commit jobs? That would help catch these issues sooner. I'm not sure how supported they're intended to be. I use them myself, but that's mainly a path of least resistance thing.

systemed:
> Sort of related: would you be open to a PR to add the Makefile builds to the commit jobs?

Yep, that would be good!

Also tangentially related... I was thinking about how we check for failed Windows builds like #661. At present we try to generate an .mbtiles of Liechtenstein, but we don't actually check to see whether it's been successfully generated or not.

cldellow:
[...]

PR #700, opened 2024-03-26
remove FAT_TILE_INDEX

cldellow:
This fixes https://github.com/systemed/tilemaker/discussions/682. Opening as a draft, as I've only done minimal testing so far.

The gist is: TileDataSource now operates in terms of `indexZoom`, not `baseZoom`. It's the lesser of `14` or your configured base zoom.

`TileCoordinatesSet` has been turned into an interface, with two implementations:

- `PreciseTileCoordinatesSet`, which is the old implementation, and used for z0-z14
- `LossyTileCoordinatesSet`, which is used for z15+. It wraps a `PreciseTileCoordinatesSet` to answer whether a tile might have objects at z15+. It gives false positives, but never false negatives. False positives are OK, as they get clipped out when writing anyway.

Some smoke testing makes me think this is working. I'll ask the reporter to give it a whirl, too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,6 +100,7 @@ file(GLOB tilemaker_src_files
 	src/sorted_node_store.cpp
 	src/sorted_way_store.cpp
 	src/tag_map.cpp
+	src/tile_coordinates_set.cpp
 	src/tile_data.cpp
 	src/tilemaker.cpp
 	src/tile_worker.cpp
--- a/Makefile
+++ b/Makefile
@@ -114,6 +114,7 @@ tilemaker: \
 	src/sorted_node_store.o \
 	src/sorted_way_store.o \
 	src/tag_map.o \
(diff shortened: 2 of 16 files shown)


cldellow:
I think this is ready for review now. @magicianlim reports that it resolves their issue with z16.

I also generated a subregion of Ontario using the OpenMapTiles profile, and compared it using the script at https://github.com/systemed/tilemaker/issues/664#issuecomment-1922733038

It's almost identical. Out of ~4,000 tiles, a single tile (z11/568/746) differs. The tile has an extra wood polygon feature. They look visually the same, and this tile is on the boundary of the clipping region for the input, so I'm assuming this is just an edge case that can be safely ignored -- e.g. perhaps something that previously got unioned into a multipolygon is now a standalone polygon.

Before:
Selection_666


After:
Selection_665

systemed:
Thank you! Seems to work well.

Issue #702, opened 2024-04-13
build problems


dieterdreist:
I got this error:
```

[  2%] Building CXX object CMakeFiles/tilemaker.dir/src/attribute_store.cpp.o
[  5%] Building CXX object CMakeFiles/tilemaker.dir/src/coordinates.cpp.o
[  7%] Building CXX object CMakeFiles/tilemaker.dir/src/coordinates_geom.cpp.o
[ 10%] Building C object CMakeFiles/tilemaker.dir/src/external/streamvbyte_decode.c.o
[ 12%] Building C object CMakeFiles/tilemaker.dir/src/external/streamvbyte_encode.c.o
[ 15%] Building C object CMakeFiles/tilemaker.dir/src/external/streamvbyte_zigzag.c.o
[ 17%] Building CXX object CMakeFiles/tilemaker.dir/src/geojson_processor.cpp.o
[ 20%] Building CXX object CMakeFiles/tilemaker.dir/src/geom.cpp.o
[ 23%] Building CXX object CMakeFiles/tilemaker.dir/src/helpers.cpp.o
/home/martin/src/tilemaker/src/helpers.cpp:167:16: error: variable has incomplete type 'struct stat64'
        struct stat64 statBuf;
                      ^
/home/martin/src/tilemaker/src/helpers.cpp:167:9: note: forward declaration of 'stat64'
        struct stat64 statBuf;
               ^
/home/martin/src/tilemaker/src/helpers.cpp:168:6: error: no viable conversion from 'stat64' to 'int'
        int rc = stat64(filename.c_str(), &statBuf);
            ^    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.
*** Error code 1

Stop.
make[2]: stopped in /usr/home/martin/src/tilemaker/build
*** Error code 1

Stop.
make[1]: stopped in /usr/home/martin/src/tilemaker/build
*** Error code 1

Stop.
make: stopped in /usr/home/martin/src/tilemaker/build

```

systemed:
Environment, OS, dependency versions etc.?

Issue #694, opened 2024-03-07
lua missing, ubuntu 23.04


paapu88:
I had to add 
```
 sudo apt-get install lua5.3 lua5.3-dev
```

To get make going in 23.04.
I could not find that in the documention

Terveisin, Markus

StephenAtty:
I had a similar problem - the install notes say to install  liblua5.1-0-dev but the run times don't get installed. It's something to do with the ubuntu packaging of lua

https://github.com/systemed/tilemaker/issues/650#issuecomment-1902259355

Issue #444, opened 2022-10-28
make not working: undefined reference to `luaL_traceback' collect2: error: ld returned 1 exit status make: *** [Makefile:90: tilemaker] Error 1


Herzbergh:
Hi, i have some trouble compiling tilemaker:

/usr/bin/ld: src/osm_lua_processing.o: in function `kaguya::lua_type_traits<kaguya::FunctionInvokerType<std::tuple<void (OsmLuaProcessing::*)(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&), void (OsmLuaProcessing::*)(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, char)> >, void>::invoke(lua_State*)':
osm_lua_processing.cpp:(.text._ZN6kaguya15lua_type_traitsINS_19FunctionInvokerTypeISt5tupleIJM16OsmLuaProcessingFvRKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESB_EMS3_FvSB_SB_cEEEEEvE6invokeEP9lua_State[_ZN6kaguya15lua_type_traitsINS_19FunctionInvokerTypeISt5tupleIJM16OsmLuaProcessingFvRKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESB_EMS3_FvSB_SB_cEEEEEvE6invokeEP9lua_State]+0x6e): undefined reference to `luaL_traceback'
/usr/bin/ld: osm_lua_processing.cpp:(.text._ZN6kaguya15lua_type_traitsINS_19FunctionInvokerTypeISt5tupleIJM16OsmLuaProcessingFvRKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESB_EMS3_FvSB_SB_cEEEEEvE6invokeEP9lua_State[_ZN6kaguya15lua_type_traitsINS_19FunctionInvokerTypeISt5tupleIJM16OsmLuaProcessingFvRKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESB_EMS3_FvSB_SB_cEEEEEvE6invokeEP9lua_State]+0x3e2): undefined reference to `luaL_traceback'
[...]

PR #713, opened 2024-04-30
Improves docker setup for basic usage to get started more easily

daniel-j-h:
Hi! :wave: Finally giving your project a try and because it's 2024 ain't nobody got time for building C++ from source, I checked out the docker images. In this changeset I make various improvements to the docker setup
1. Using the stable debian bookworm as a base image as bullseye is outdated
2. Not installing the development libraries with the tilemaker binary as they're only needed for building
3. Compiling in parallel so that the docker build doesn't take 10+ minutes to build
4. Adding a default command of `--help` so that if folks do `docker run` they get to see the tilemaker help
5. Updated the docker readme and added a usage section in the default readme to use docker by default

I would highly recommend defaulting to telling users to just use the docker image and not having to compile tilemaker from source. There are so many compilation problems caused by dependencies to install, compiler and linker versions, and so on.

Just from a first look
- https://github.com/systemed/tilemaker/issues/702
- https://github.com/systemed/tilemaker/issues/694
- https://github.com/systemed/tilemaker/issues/444

If we could dumb down the getting started instructions to

    docker run .. tilemaker input.pbf --output output.pbf

that would make it easier for users and hopefully less likely for us to get support tickets.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,52 +1,56 @@
-FROM debian:bullseye-slim AS src
+FROM debian:bookworm-slim AS src
 LABEL Description="Tilemaker" Version="1.4.0"
 
-ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+    build-essential \
+    liblua5.1-0-dev \
+    libsqlite3-dev \
+    libshp-dev \
+    libboost-program-options-dev \
+    libboost-filesystem-dev \
+    libboost-system-dev \
+    libboost-iostreams-dev \
+    rapidjson-dev \
+    cmake \
+    zlib1g-dev && \
+    rm -rf /var/lib/apt/lists/*
 
-# install dependencies
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends \
-      build-essential \
-      liblua5.1-0 \
-      liblua5.1-0-dev \
-      libsqlite3-dev \
-      shapelib \
-      libshp-dev \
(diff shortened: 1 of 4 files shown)


systemed:
This is great - thank you!

I know practically nothing about docker - the support that we have so far is entirely down to other contributors. So I'm very happy to merge this but I may be leaning on you and others for support in the future.

> it's 2024 ain't nobody got time for building C++ from source

Call me old-school 😉 (but also it's seriously fast on Apple silicon...).

daniel-j-h:
Got it, I believe it's fair to say that Docker will not go away so can recommend looking into it.

In the context of native projects like tilemaker the benefit I see
- on the dev side is being able to compile in a self-contained development environment with a specific compiler against one set of libraries, so you don't have to support the combination of gcc, clang x library versions but e.g. whatever is in Debian LTS
- on the user side the benefit is you don't have to compile anything at all but can just run the pre-built docker image

If you want to give this PR a try locally

```bash
docker build -t tilemaker .
docker run -it --rm tilemaker
```

I just rebased against master and added the tilemaker binary to PATH so https://github.com/systemed/tilemaker/pull/465 can be closed, too.

Like a said in https://github.com/systemed/tilemaker/pull/713#issuecomment-2086138281 in the future a second iteration would be publishing not just a single tilemaker docker image but tagging them e.g. by date or version or commit hash. This can be a follow-up pull request.

Happy to get this in?

leonardehrenfried:
Now that this one is also merged, would it be possible to take a look at https://github.com/systemed/tilemaker/pull/685 ?

PR #721, opened 2024-05-14
Update RUNNING.md docker old config name

mvanlaar:
Replace the old config name with the new .json extension.


systemed:
Thank you!

Issue #720, opened 2024-05-13
Ocean and land area is missing


BinoyKadakkal:
Ocean and land area is missing while converting .osm.pbf file to .mbtiles. and shows the error 

_### Unable to open coastline/water_polygons.shp or coastline/water_polygons.SHP.
Reading shapefile urban_areas
Unable to open landcover/ne_10m_urban_areas/ne_10m_urban_areas.shp or landcover/ne_10m_urban_areas/ne_10m_urban_areas.SHP.
Reading shapefile ice_shelf
Unable to open landcover/ne_10m_antarctic_ice_shelves_polys/ne_10m_antarctic_ice_shelves_polys.shp or landcover/ne_10m_antarctic_ice_shelves_polys/ne_10m_antarctic_ice_shelves_polys.SHP.
Reading shapefile glacier
Unable to open landcover/ne_10m_glaciated_areas/ne_10m_glaciated_areas.shp or landcover/ne_10m_glaciated_areas/ne_10m_glaciated_areas.SHP._
![Screenshot from 2024-05-13 16-12-04](https://github.com/systemed/tilemaker/assets/108899419/04ae8d13-8c79-4b16-b5b9-01ec05b94f34)

Tilemaker is running in docker

systemed:
From the readme:

> To include sea tiles, create a directory called coastline in the same place you're running tilemaker from, and then save the files from https://osmdata.openstreetmap.de/download/water-polygons-split-4326.zip in it, **such that tilemaker can find a file at coastline/water_polygons.shp.**

The error you are getting shows that you don't have a file at the location where tilemaker is expecting to find it. Make sure that you don't have any extra levels of directory in there.

(Please use Github discussions, not issues, for help questions. Thank you :) )

systemed:
Ah, you're using Docker. Sorry, I don't know how the filesystem works there, hopefully someone else will.

baurzhan:
Run container without --rm option. Copy folder to container and restart container:
` docker cp landcover tilemaker:/usr/src/app/`
`docker cp coastline tilemaker:/usr/src/app/`
`docker start tilemaker`

Where "tilemaker" is the name of your container

BinoyKadakkal:
it still shows the same problem

systemed:
@daniel-j-h Could I trouble you for any thoughts on this?

daniel-j-h:
I believe this has to do with absolute vs relative paths; in your `config.json` can you check if you have paths such as
[...]

PR #723, opened 2024-05-15
Adds convenience scripts to download coastline and landuse data

daniel-j-h:
Splitting off from #720 - this changeset adds my convenience scripts to download coastline and landcover data. It's inconvenient to go to these websites, manually find the dataset needed, download and unpack it, and so on.

Let me know what a good file structure is, where we should put them, if I should consolidate them into a single script, and so on.

Adding a section to the readme; preview:

https://github.com/daniel-j-h/tilemaker/tree/add-coastline-landcover-download?tab=readme-ov-file#coastline-and-landcover

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,16 +50,23 @@ tilemaker keeps everything in RAM by default. To process large areas without run
 
     tilemaker /path/to/your/input.osm.pbf /path/to/your/output.mbtiles --store /path/to/your/ssd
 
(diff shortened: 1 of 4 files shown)


daniel-j-h:
Adding here that in the past I have also used

https://daylightmap.org/2023/10/11/landcover.html

but long term this seems to get folded into the Overture distribution.

systemed:
Superb, thank you. Happy to go with this as it is!

PR #727, opened 2024-06-04
Explicitly added lua5.1 to the install list

SomeoneElseOSM:
On an "out of the box" Debian 12 server I noticed that lua5.1 was needed in addition to liblua5.1.0-dev (if using lua 5.1). Those packages are also valid on Ubuntu 22.04.

Alternatively, "... lua5.3 liblua5.3-dev ..." also seems to work (at least on Ubuntu 22.04).


laem:
I was about to make the same PR

systemed:
Thanks!

Issue #751, opened 2024-09-09
Feature: Add BUILD_DEBUG argument when building Docker image


oobayly:
When creating #750 I needed to be enable debugging for the docker image. Passing an optional `BUILD_DEBUG` argument with a non-empty value:
* Runs cmake with the `-DCMAKE_BUILD_TYPE=Debug` flag
* Doesn't strip the executables after they're build
* Installs gdb

`docker build --build-arg BUILD_DEBUG=1 --tag  tilemaker/debug .`

PR #753, opened 2024-09-09
Feature/751 Add optional BUILD_DEBUG Docker build argument

oobayly:
See systemed/tilemaker#751

Makes it easier to enable debugging for the docker image. Passing an optional BUILD_DEBUG argument with a non-empty value:

* Runs cmake with the -DCMAKE_BUILD_TYPE=Debug flag
* Doesn't strip the executables after they're build
* Installs gdb

`docker build --build-arg BUILD_DEBUG=1 --tag  tilemaker/debug .`


systemed:
Thank you!

Issue #752, opened 2024-09-09
Feature: Add --quiet option to tilemaker executable


oobayly:
When creating #750 I had issues getting the VS code debugger to correctly (probably related to Microsoft/vscode-cpptools#1385. Disabling the large amount of STDOUT ouput generated when running `tilemaker` in debug solved the issue.

I've added an optional `--quiet` flag to `tilemaker` that defaults to false. When set to true:
```cpp
if (options.quiet) {
  // Suppress anything written to std out
  std::cout.setstate(std::ios_base::failbit);
}
```

I'm not sure that this is the best/correct way to do this, but it works. I'm happy to take any advice if it should be changed.

PR #754, opened 2024-09-09
Feature/752 Add optional --quiet flag to tilemaker

oobayly:
See systemed/tilemaker#752

I had issues getting the VS code debugger to correctly (probably related to https://github.com/microsoft/vscode-cpptools/issues/1385. Disabling the large amount of STDOUT ouput generated when running tilemaker in debug solved the issue.

I've added an optional --quiet flag to tilemaker that defaults to false. When set to true:

```c++
if (options.quiet) {
  // Suppress anything written to std out
  std::cout.setstate(std::ios_base::failbit);
}
```

I'm not sure that this is the best/correct way to do this, but it works. I'm happy to take any advice if it should be changed.


systemed:
Thanks!

Issue #748, opened 2024-09-06
Add global lua function to list all tags


oobayly:
Could we have a global function that would allow us to iterate through all the key-value pairs in the `currentPostScanTags`? I would to be able to add attributes which have a certain prefix.

I'm trying to create a layer whose tags can't easily be abstracted, and would like to be able to include all the tags (at the cost of increased tile size) that may be needed for rendering via MapLibre without having to rebuild the tiles.

oobayly:
I've create a patch for this, but my C++ isn't great, so while it does work, there are probably better methods. I'll leave this up here for any comments, and I can create a PR is desired.

```diff
diff --git a/include/osm_lua_processing.h b/include/osm_lua_processing.h
index f4d664d..ca7627a 100644
--- a/include/osm_lua_processing.h
+++ b/include/osm_lua_processing.h
@@ -111,6 +111,9 @@ public:
 	// Get the ID of the current object
 	std::string Id() const;
 
+	// Gets a table of all the OSM tags
+	kaguya::LuaTable AllTags(kaguya::State& luaState);
+
 	// Check if there's a value for a given key
 	bool Holds(const std::string& key) const;
 
diff --git a/src/osm_lua_processing.cpp b/src/osm_lua_processing.cpp
index 545b1f3..e2ea791 100644
--- a/src/osm_lua_processing.cpp
+++ b/src/osm_lua_processing.cpp
@@ -118,6 +118,20 @@ template<>  struct kaguya::lua_type_traits<protozero::data_view> {
 };
 
 std::string rawId() { return osmLuaProcessing->Id(); }
+kaguya::LuaTable rawAllTags() {
+	if (osmLuaProcessing->isPostScanRelation) {
+		return osmLuaProcessing->AllTags(*g_luaState);
+	}
+
+	auto tags = osmLuaProcessing->currentTags->exportToBoostMap();
+	kaguya::LuaTable tagsTable = g_luaState->newTable();
+
+	for (const auto& kv : tags) {
+		tagsTable[kv.first] = kv.second;
+	}
+
+	return tagsTable;
+}
 bool rawHolds(const KnownTagKey& key) {
 	if (osmLuaProcessing->isPostScanRelation) {
 		return osmLuaProcessing->Holds(key.stringValue);
@@ -196,6 +210,7 @@ OsmLuaProcessing::OsmLuaProcessing(
 
 	osmLuaProcessing = this;
 	luaState["Id"] = &rawId;
[...]

PR #755, opened 2024-09-09
Feature/748 Add AllKeys and AllTags Lua functions

oobayly:
See systemed/tilemaker#748

This adds two exported functions for Lua:
* AllKeys - returns an array of all the keys for the OSM tags
* AllTags - returns a table of all the OSM tags

This allows us to add all the OSM nodes that match a pattern, have a desired prefix, etc without need to know in advance the actual tags.

I have this working on my tilemake fork and it's working quite nicely. I'm not a C++ guy, but I've tried duplicating the code style.


systemed:
This works really well - thank you! (Sorry for the delay in merging, just catching up with things.)

PR #757, opened 2024-09-13
docs: adjust role explanation for NextRelation

cldellow:
The current docs describe a single return value that is an array of size 2, but the `NextRelation()` function actually has two return values.

See https://www.lua.org/pil/5.1.html for more.

Hat tip to @BenOnTrack, who noticed it in https://github.com/systemed/tilemaker/discussions/749


systemed:
Thank you!

Issue #747, opened 2024-09-04
Fails to build with Boost 1.86.0


wdaus:
I'm not very familiar with C++ or Boost, but it looks like `mmap_allocator.cpp` is trying to use a method that no longer exists in Boost 1.86.0 (`destroy` in `interprocess::allocator`).

```
make
...
g++ -O3 -Wall -Wno-unknown-pragmas -Wno-sign-compare -std=c++14 -pthread -fPIE -DTM_VERSION=v3.0.0
-o src/mmap_allocator.o -c src/mmap_allocator.cpp -I/usr/local/include -isystem ./include -I./src -I/usr/include/luajit-2.1 DLUAJIT
src/mmap_allocator.cpp: In static member function ‘static void void_mmap_allocator::destroy(void*)’:
src/mmap_allocator.cpp:227:42: error: ‘using allocator_t = class boost::interprocess::allocator<unsigned char,
boost::interprocess::segment_manager<char, boost::interprocess::rbtree_best_fit<boost::interprocess::null_mutex_family>,
boost::interprocess::iset_index> >’ {aka ‘class boost::interprocess::allocator<unsigned char,
boost::interprocess::segment_manager<char, boost::interprocess::rbtree_best_fit<boost::interprocess::null_mutex_family>,
boost::interprocess::iset_index> >’} has no member named ‘destroy’
  227 |                         return allocator.destroy(reinterpret_cast<uint8_t *>(p));
```

systemed:
https://github.com/boostorg/interprocess/commit/929b488e6c847c729a53654b26a82a5a5f513b15 looks like the change in question.

cldellow:
Oops, I didn't put the magic words in the PR, so GitHub didn't autoclose this issue.

@wdaus I think this has been fixed in #759 - can you give that a whirl and close the issue if it works for you now?

wdaus:
@cldellow thanks for the quick fix. Can confirm that build succeeds on Arch Linux with the latest boost version.

```Name            : boost
Version         : 1.86.0-2
Description     : Free peer-reviewed portable C++ source libraries (development headers)
Architecture    : x86_64
...
Build Date      : Sat 07 Sep 2024 04:51:12 AM AEST
...```

wdaus:
Closing issue

PR #758, opened 2024-09-13
attempt to fix workflow failures

cldellow:
GitHub has deprecated the v2 version of the upload-artifact action: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

Workflows automatically fail now due to using the outdated version - hence the failures on [this documentation-only PR](https://github.com/systemed/tilemaker/pull/757)

I didn't actually read the release notes, I'm just YOLOing into v4 and hoping that works. If it still fails, I'll investigate more.


cldellow:
D'oh, the workflow runs now, but fails due to #747. Reading the current docs for upload-artifact@v4, I think the YAML interface is the same, and so I expect it's safe to merge this.

cldellow:
Closing in favour of #759

PR #760, opened 2024-09-21
Fix #750: allow no more than 512 attribute names

### LINKED ISSUE #750: Bug: SegFault caused by invalid string
I came across this when adding a large number of attributes to a nodes. In my case I was adding all the OSM tags with a `railway:` prefix (I'm not planning on doing always doing this, but it's useful for development). This was fine for a small region (Köln Rebenz), but would end with SIGSEGV when using the parent area (Nordrehin Westfalen).

My first thought was that it's an OOM error, but this turned out not to be the case:
* First, I used the `--store` flag, but this made no difference.
* The total memory being used was only 3.5GB. Docker has up to 32GB on my machine with a total of 64GB physical memory.
* Increasing the number of nodes greatly (by processing the entirety of Germany), but with only subsets of the `railways:` tags did work, with a similar 3.5GB usage.
* At one stage (I can't remember how), I managed to get a readable error that suggested a string issue.

So, it seems to be an issue with attributes, I:
* Rebuilt tilemaker with debug and didn't strip the symbols
* Enabled coredumps in docker and installed gdb
* Made sure to run with only a single thread
* Managed obtain a backtrace for the offending error - attached [tilemaker-backtrace.txt](https://github.com/user-attachments/files/16930540/tilemaker-backtrace.txt)
* In my case there are 301 items in the `AttributeStore->keyStore`. *I appreciate that this many not be great practice, but nor should this be causing a SegFault*.

In short, it was **`error reading variable: Cannot access memory at address 0x1a149fe`**, and was been thrown in [OutputObject::writeAttributes:49](https://github.com/systemed/tilemaker/blob/eab08d189ad97ddf5db7d915bcabe42ad3dab6af/src/output_object.cpp#L49)

Attaching a debugger, I found that address of the string `const std::string& key = attributeStore.keyStore.getKeyUnsafe(it->keyIndex);` was invalid, and attempting to read it is causing an access violation. More specifically, the iterator `it` had an invalid (negative) `keyIndex` value.
image
What's interesting is that the other properties for the `AttributePair` is valid, `valueType` and the string in `stringValue`.

I'll try setting up a simple (but completely unrealistic) set up so it can be reproduced on demand.

As an aside, I'm going to create another 2 issues with associated PR that helped me with getting live debugging to work via docker.
* #751 
* #752

cldellow:
This fixes two issues:

- use an unsigned type, so we can use the whole 9 bits and have 512 keys, not 256
- fix the bounds check in AttributeKeyStore to reflect the lower threshold that was introduced in #618

Hat tip @oobayly for reporting this.

Fixes #750.

### DIFF
--- a/include/attribute_store.h
+++ b/include/attribute_store.h
@@ -46,7 +46,7 @@ enum class AttributePairType: uint8_t { String = 0, Float = 1, Bool = 2 };
 // AttributePair is a key/value pair (with minzoom)
 #pragma pack(push, 1)
 struct AttributePair {
-	short keyIndex : 9;
+	unsigned short keyIndex : 9;
 	AttributePairType valueType : 2;
 	uint8_t minzoom : 5; // Support zooms from 0..31. In practice, we expect z16 to be the biggest minzoom.
 	union {
--- a/src/attribute_store.cpp
+++ b/src/attribute_store.cpp
@@ -35,8 +35,8 @@ uint16_t AttributeKeyStore::key2index(const std::string& key) {
 	uint16_t newIndex = keys.size();
 
 	// This is very unlikely. We expect more like 50-100 keys.
-	if (newIndex >= 65535)
-		throw std::out_of_range("more than 65,536 unique keys");
+	if (newIndex >= 512)
+		throw std::out_of_range("more than 512 unique keys");
 
 	keys2indexSize++;
 	keys.push_back(key);
--- a/test/attribute_store.test.cpp
+++ b/test/attribute_store.test.cpp
@@ -89,13 +89,38 @@ MU_TEST(test_attribute_store_reuses) {
 
 		mu_check(s1aIndex == s1bIndex);
 	}
+}
 
+MU_TEST(test_attribute_store_capacity) {
+	// We support a maximum of 511 attribute name, so confirm that we can roundtrip
+	// this value.
+	AttributePair pair(511, true, 0);
+	mu_check(pair.keyIndex == 511);
 
+	// Confirm that the attribute store will throw if we try to add more than 511 keys.
+	AttributeKeyStore keys;
+
+	for (int i = 1; i <= 511; i++) {
+		const uint16_t keyIndex = keys.key2index("key" + std::to_string(i));
+		mu_check(keyIndex == i);
+	}
+
+	// Trying to add a 512th key should throw
+	bool caughtException = false;
+
+	try {
+		keys.key2index("key512");
+	} catch (std::out_of_range) {
+		caughtException = true;
+	}
+
+	mu_check(caughtException == true);
 }
 
 MU_TEST_SUITE(test_suite_attribute_store) {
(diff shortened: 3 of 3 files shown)


cldellow:
Hmmm. The Ubuntu Makefile build has failed twice in a row, on the "Generate mbtiles with Github action" workflow step. It looks like a segfault happens:

```
2024-09-21T03:19:08.3604627Z Reading .pbf liechtenstein.osm.pbf
2024-09-21T03:19:08.3967057Z 
2024-09-21T03:19:08.3970871Z (Scanning for ways used in relations: 0%)           (23 ms)
2024-09-21T03:19:08.4895426Z /home/runner/work/_temp/b7889b5c-995f-44e1-8026-c41c35ac1e82.sh: line 3:  2472 Segmentation fault      (core dumped) ./tilemaker liechtenstein.osm.pbf --config=resources/config-openmaptiles.json --process=resources/process-openmaptiles.lua --output=liechtenstein.mbtiles --verbose --store /tmp/store
2024-09-21T03:19:08.4906982Z ##[error]Process completed with exit code 139.
```

That's pretty coincidental! I think the point of that workflow step is to verify that https://github.com/systemed/tilemaker/blob/master/action.yml is a valid GitHub action file, in the event that the PR has modified it and broken it, for example.

What's odd is that this PR doesn't change that file, and the action itself uses the `docker://ghcr.io/systemed/tilemaker:master` image, which (a) isn't an artifact of this PR--it's the previously built image from the last push to master and (b) ought to be the same image on the Ubuntu Makefile and Ubuntu CMake builds, so it's odd that the Makefile build failed twice in a row, and the CMake build worked twice in a row. Maybe just random luck.

systemed:
Thank you!

The CI seems to have become very flakey recently, I think largely due to dependency issues I don't fully understand. It needs sorting at some point but I don't think erratic partial failures are currently a reason not to merge stuff.

cldellow:
:+1: Yup, not blocking merges makes a lot of sense.

I think this specific failure--a segfault running a previously-published docker image--hints at a bug in tilemaker vs a bug in the CI scripts, unfortunately. (Separately, the CI may also be flakey, of course!)

I did some futzing locally and it looks like I'm able to repro if I run tilemaker in a loop under gdb (following the instructions at https://stackoverflow.com/a/6546674/1011680). For me, it seems to occur about 1% of the time. It occurs with or without luajit, and with or without store mode. I haven't yet seen it happen with `--threads 1`, so I think we're probably not protecting a data structure appropriately against concurrent access, causing undefined behaviour. The stack traces in the faulting thread are all over the place, which I think supports this -- something read/wrote into a spot it shouldn't have, corrupted whatever happened to be nearby, and then the world ends in a random location shortly thereafter when something steps on the landmine that just got created.

The corruption must happen relatively early, which is handy. I hacked tilemaker just to do the read node phase, and then exit, so that I could get more iterations of the test. I'll keep poking around and see if I can figure out what's going on.

PR #764, opened 2024-09-29
Push Docker image on all pull requests

cldellow:
Previously, we pushed a docker image only when building `master`.

Instead, also push it for all PRs.

My fig leaf motivation: this makes it easy for someone to test out a PR, as they can just update their docker command to reference `ghcr.io/systemed/tilemaker:pr-1234` or whatever.

My real motivation is selfish. In my fork of tilemaker, I occasionally have patches that don't exist in upstream. I'd like to just maintain a PR against my fork (eg, this one: https://github.com/cldellow/tilemaker/pull/4), and let GitHub build a docker image and push it to ghcr.io for use in my pipeline.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,17 +178,8 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
 
-      - name: Build Docker image
-        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
-        if: ${{ github.ref != 'refs/heads/master'}}
-        with:
-          context: .
-          push: false
-          tags: ${{ steps.meta.outputs.tags }}
-          labels: ${{ steps.meta.outputs.labels }}
       - name: Build and push Docker image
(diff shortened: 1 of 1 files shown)


cldellow:
D'oh, the docker build fails with:

```
#23 ERROR: denied: installation not allowed to Write organization package
------
 > pushing ghcr.io/systemed/tilemaker:pr-764 with docker:
------
ERROR: denied: installation not allowed to Write organization package
```

I guess enabling this might require some changes on your end to tilemaker's GitHub Actions security settings. That makes this a scarier change than I was hoping for.

Would you be open to that? If not, it's not a big deal - I can just bundle this patch into the other patches in my fork.

I think the relevant knob is at https://github.com/users/systemed/packages/container/tilemaker/settings -- I _think_ that would grant GitHub Actions running as people other than you the right to push into the systemed/tilemaker namespace on ghcr.io. That's a little risky, of course: a rogue person could, say, gin up a PR that puts something malicious in the master docker image.

It's mitigated a bit by the fact that (a) such tampering would be noisy and easily discovered and (b) the tilemaker repo is set up such that unknown contributors need approval from you first before they can run workflows.

cldellow:
Given that this requires some security risks on your end + that it was mostly driven by my own laziness, it's hard to justify.

PR #766, opened 2024-10-04
docker: luarocks, luaflock, lua-sql-sqlite3

cldellow:
This updates the Docker container to have some useful packages that a user might want available to their Lua scripts:

- `luaflock`: BSD-style flock
- `lua-sql-sqlite3`: SQLite bindings for Lua

It also updates tilemaker's Lua error handler to print the lua error code and error message string.

It seems like not all Lua errors have stack traces/tracebacks -- for example, requiring a non-existent module:

before this change:

```
lua runtime error:
terminate called after throwing an instance of 'OsmLuaProcessing::luaProcessingException'
  what():  std::exception
```

after this change:

```
lua runtime error 2:
./file_append.lua:7: module 'flock' not found:
	no field package.preload['flock']
	no file './flock.lua'
	no file '/usr/local/share/lua/5.1/flock.lua'
	no file '/usr/local/share/lua/5.1/flock/init.lua'
	no file '/usr/local/lib/lua/5.1/flock.lua'
	no file '/usr/local/lib/lua/5.1/flock/init.lua'
	no file '/usr/share/lua/5.1/flock.lua'
	no file '/usr/share/lua/5.1/flock/init.lua'
	no file './flock.so'
	no file '/usr/local/lib/lua/5.1/flock.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/5.1/flock.so'
	no file '/usr/lib/lua/5.1/flock.so'
	no file '/usr/local/lib/lua/5.1/loadall.so'
terminate called after throwing an instance of 'OsmLuaProcessing::luaProcessingException'
  what():  std::exception
```

A drawback of this is that the case where there _was_ already a useful error message is now quite spammy, e.g.:

```
lua runtime error 2:
/home/cldellow/src/basemap/export/city-parks.lua:27: attempt to perform arithmetic on global 'w' (a nil value)
table  `/home/cldellow/src/basemap/export/city-parks.lua:27: attempt to perform arithmetic on global 'w' (a nil value)'  `/home/cldellow/src/basemap/export/city-parks.lua:27: attempt to perform arithmetic on global 'w' (a nil value)'  
Lua error on node 21760675
```

...that's kind of unfortunate, but for me, is still a good tradeoff vs having nothing in some cases.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,9 +9,11 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-ins
     libboost-program-options-dev \
     libboost-filesystem-dev \
     libboost-system-dev \
+    luarocks \
     rapidjson-dev \
     cmake && \
-    rm -rf /var/lib/apt/lists/*
+    rm -rf /var/lib/apt/lists/* && \
+    luarocks install luaflock
 
 WORKDIR /usr/src/app
 
@@ -35,6 +37,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-ins
     liblua5.1-0 \
     shapelib \
     libsqlite3-0 \
+    lua-sql-sqlite3 \
     libboost-filesystem1.74.0 \
     libboost-program-options1.74.0 && \
     rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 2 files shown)


systemed:
Thank you! This now has a conflict on the Dockerfile as a result of merging other PRs - I'm sure it's pretty straightforward but Docker is not really my thing so I'm a little anxious about breaking stuff; could you take a look?

cldellow:
Will do.

systemed:
Thank you!

PR #770, opened 2024-10-06
don't require any Lua functions

### LINKED ISSUE #142: Don't die if Lua functions missing
It's plausible that the user might not need Lua functions, if they're just making tiles from shapefiles without any OSM data. Currently we bomb out with a kaguya "Type mismatch" if the Lua profile is empty.

cldellow:
`node_function` and `way_function` are now optional.

Fixes #142.


systemed:
Thank you!

PR #795, opened 2025-01-30
pmtiles: always set leafStart when closing file [#794]

### LINKED ISSUE #794: root-directory-only pmtiles archives have incorrect header offsets
## Reproduce

Run `tilemaker monaco.osm.pbf --output monaco.pmtiles` and the view the result metadata on [pmtiles.io](https://pmtiles.io)

```
root directory: offset=127 len=157
metadata: offset=254954 len=518
leaf directories: offset=0 len=254954
tile data: offset=16384 len=18446744073709535000
```

The leaf directory offset/length is incorrect, as well as the tile data length. Other fields look OK. 
 
* this does not affect the ability to view the tiles, because the `leaf_directory_offset` and `tile_data_length` are not used by readers.
* It does affect the ability to `pmtiles edit` to modify the JSON metadata using the [pmtiles CLI](https://github.com/protomaps/go-pmtiles) and is also necessary for `pmtiles cluster` to work, since both these operations rely on that information being correct to efficiently make a copy of the file using few I/O operations.

In [pmtiles cli v1.25](https://github.com/protomaps/go-pmtiles/releases/tag/v1.25.0) there are additional checks to detect this case.

(PR coming)

bdon:
* the `leafStart` was only assigned in flushEntries, which never happens if there is only a root directory.
* resulted in erroneous leafDirectoryOffset and tileDataLength in the pmtiles header.

Detect with `pmtiles verify`:

```
pmtiles verify main.pmtiles
2025/01/30 15:11:57 main.go:223: Failed to verify archive, Leaf directories offset=0 must not be 0
exit status 1
```

Fixes #794

### DIFF
--- a/src/pmtiles.cpp
+++ b/src/pmtiles.cpp
@@ -28,6 +28,7 @@ void PMTiles::close(std::string &metadata) {
 	// add all tiles to directories, writing leaf directories as we go
 	std::vector<pmtiles::entryv3> rootEntries;
 	std::vector<pmtiles::entryv3> entries;
+	leafStart = outputStream.tellp();
 	if (isSparse) {
 		for (auto it : sparseIndex) {
 			appendTileEntry(it.first, it.second, rootEntries, entries);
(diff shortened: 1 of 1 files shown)


systemed:
Thank you!

PR #808, opened 2025-04-07
fix build on non-glibc Linux

### LINKED ISSUE #807: Build fails on non-glibc Linux due to use of stat64
helpers.cpp currently uses `stat64` on all Linux systems, causing the build to fail on non-glibc based systems such as Alpine Linux which do not have LFS64.

edaigle:
* non-glibc Linux do not have stat64, causing build to fail on e.g. musl/Alpine.
* this commit adds a check to make sure we are on a glibc system before using stat64.

Closes #807

### DIFF
--- a/src/helpers.cpp
+++ b/src/helpers.cpp
@@ -16,7 +16,7 @@
 #define stat64 __stat64
 #endif
 
-#if defined(__APPLE__)
+#if defined(__APPLE__) || !defined (__GLIBC__)
 #define stat64 stat
 #endif
 


src/helpers.cpp
```diff
@@ -16,7 +16,7 @@
 #define stat64 __stat64
 #endif
 
-#if defined(__APPLE__)
+#if defined(__APPLE__) || !defined (__GLIBC__)
```
cmeerw, line 19: I don't think that change is correct - it would conflict with the define for `stat64` when `_MSC_VER` is defined (above)

Maybe (assuming that `__GLIBC__` isn't defined for `__APPLE__` anyway):

```
#ifdef _MSC_VER
#define stat64 __stat64
#elif !defined(__GLIBC__)
#define stat64 stat
#endif
```

PR #818, opened 2025-04-16
feat: write attribution in pmtiles metadata

echozio:
This patch includes the attribution in the PMTiles metadata if it exists in config.json.
This should allow #782 to function with PMTiles, too, not just MBTiles.


systemed:
Thank you (and sorry for the delay merging it)!

PR #825, opened 2025-06-23
fix a number of bugs and differences from openmaptiles vector tile schema

arichnad:
fix a number of bugs and differences from openmaptiles vector tile schema

1.  zoom 4, 5, and 7 road values were incorrectly being used.

2.  openmaptiles shows admin 3 boundaries much higher.

3.  unpaved and paved values are parsed from surface list

4.  remove duplicate variable definitions.

5.  SetMinZoomByAreaWithLimit allows up to minzoom 4

6.  fix whitespace and a few variable issues.

7.  SetMinZoomByAreaWithLimit improvements (allow zoom level 1 and above)

8.  recalculate a few constants to use 40.075e6 meters

9.  do not set an unset name.

10.  separating out minor road values and path road values allows for a more correct interpretation of the zoom levels.

### DIFF
--- a/resources/process-openmaptiles.lua
+++ b/resources/process-openmaptiles.lua
@@ -30,6 +30,11 @@ function Set(list)
 end
 
 -- Meters per pixel if tile is 256x256
+ZRES0 = 156542.97
+ZRES1  = 78271.48
+ZRES2  = 39135.74
+ZRES3  = 19567.87
+ZRES4  = 9783.94
 ZRES5  = 4891.97
 ZRES6  = 2445.98
 ZRES7  = 1222.99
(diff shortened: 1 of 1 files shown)


arichnad:
7. fixes a bug with setting an empty name:  a set "empty" name will override the `ref` if one is set.

resources/process-openmaptiles.lua
```diff
@@ -826,7 +839,12 @@ end
 -- Set minimum zoom level by area but not below given minzoom
 function SetMinZoomByAreaWithLimit(minzoom)
 	local area=Area()
-	if     minzoom <= 6 and area>ZRES5^2  then MinZoom(6)
+	if     minzoom <= 1 and area>ZRES0^2  then MinZoom(1)
+	elseif minzoom <= 2 and area>ZRES1^2  then MinZoom(2)
+	elseif minzoom <= 3 and area>ZRES2^2  then MinZoom(3)
+	elseif minzoom <= 4 and area>ZRES3^2  then MinZoom(4)
+	elseif minzoom <= 5 and area>ZRES4^2  then MinZoom(5)
+	elseif minzoom <= 6 and area>ZRES5^2  then MinZoom(6)
```
M393, line 847: > 5. SetMinZoomByAreaWithLimit allows up to minzoom 4

I'm curious why does this only allow up to minzoom 4?
I'm trying to show lakes on all zoom levels. With this change they do now show up on zoom level 4 and 5, but not earlier.
E.g. https://www.openstreetmap.org/relation/945096

My config:
```json
"water":             { "minzoom": 0,  "maxzoom": 14, "simplify_below": 12, "simplify_level": 0.0003, "simplify_ratio": 2}
```

M393: Shouldn't the water now be displayed on all zoom levels if it is large enough when minzoom is set to 0?
For me it is not showing on zoom 0-3.

arichnad:
I have fixed the merge conflicts, thanks!

systemed:
Thank you and sorry for the slow merge!

PR #827, opened 2025-06-30
Add AttributeInteger for writing int values

systemed:
Currently we offer only AttributeNumeric for writing numeric values from Lua. This is mapped to a (4-byte) float in the vector tile. Floats are lossy and potentially less efficient.

This adds an `AttributeInteger` method, which writes int values to the vector tile. Internally this is currently stored as a double (so much less lossy), and on output to the vector tile it should use fewer bytes in most circumstances.


bmorg:
Thanks!

Does this need fixing in the repo or is this a problem in the data I'm using?

PR #828, opened 2025-06-30
not using visvalingam breaks earcut in maplibre.

arichnad:
earcut gets into some crazy situations if you don't use visvalingam to simplify polygons:

lakes in the United States cause earcut in maplibre to run especially long (100 times normal).


arichnad:
Are there any issues with this pull request?

systemed:
Thank you for this and sorry for the delay!

PR #832, opened 2025-07-08
document `include_ids` and `compress` in configuration readme

etienneJr:
Following #830, I propose this very simple PR to document in CONFIGURATION.md the incompatibility between `include_ids` and `combine_xxx` settings. (I don't think I'll be able to dive into the code to generate a warning when both options are used)


etienneJr:
Following #833 I also add the information that pmtiles are gzip compressed.

Issue #740, opened 2024-08-06
Keep OSM feature type in the final mbtiles file


laem:
Hi, I found the include_ids option in the parameters, but I can't find an option to store the OSM feature type as well : node, way or relation. 

It makes it very difficult to link the tiles' nodes to OSM on the final map. 

I wonder if it's possible to add someting in the process-openmaptiles.lua file... Any help appreciated, thanks :)

laem:
Exemple inspecting the mbtiles produces with the default tilemaker options with include_ids. 


![image](https://github.com/user-attachments/assets/be286210-c37e-4c51-
a9a2-8a9f178eaa98)
This "Point" is in fact a way https://www.openstreetmap.org/way/64518593#map=20/47.55544/-3.13397&layers=H

systemed:
The `$id` is a dedicated field in the vector tiles, rather than an user-set attribute: see 4.2 in https://github.com/mapbox/vector-tile-spec/tree/master/2.1. There is no dedicated field for an "id type".

If you do want to add it as an attribute, you could certainly do that in process.lua. In `node_function` you'd add a line like `Attribute("obj_type", "node")`, in `way_function` you'd add `Attribute("obj_type","way")` and so on.

laem:
Thanks for your quick answer ! 
I just tried this https://github.com/laem/gtfs/blob/master/tilemaker/resources/process-openmaptiles.lua#L124 without success yet, but I'm investigating.

laem:
Mmmh maybe WritePOI needs to be modified as well.

laem:
Awesome :) 

![image](https://github.com/user-attachments/assets/0563e43c-265c-4c8a-a41e-bc9e92f10d59)

laem:
Hi @systemed, what about relations that are handled by `way_function`, is it possible to know that a feature treated by `way_function` is an OSM relation ? 

> Tilemaker handles multipolygon relations natively. The combined geometries are processed as ways (i.e. by way_function), so if your function puts buildings in a 'buildings' layer, tilemaker will cope with this whether the building is mapped as a simple way or a multipolygon. The only difference is that they're given an artificial ID. Multipolygons are expected to have tags on the relation, not the outer way.

E.g. some schools can be ways or relations. [This one](https://api.openstreetmap.org/api/0.6/relation/5676522/full.json) is a relation, but in my tiles I want to keep track of the original OSM object, so I need to keep its original relation status. 

I haven't been able to find a way to do that. 

[...]

PR #834, opened 2025-07-12
create function to get OSM type

etienneJr:
### Context
Following #740 we use `Attribute("obj_type","way")` in process.lua `way_function` to include this information in cartes.app tiles. But it is not working well for ways that come from a relation, we have to check the tags, it is not a direct information.

### Solution
I wrote this PR to create a function `OsmType()` which returns the true OSM type of the object, on the same model than the existing `Id()` function.

### Example
Using this in process.lua:
```
Attribute("OsmId", Id())
Attribute("OsmType", OsmType())
```
I can directly get tiles with the correct type information
image


systemed:
Thanks - this looks good! Sorry for the delay in looking at it. In:

    return (isRelation ? "relation " : isWay ? "way " : "node ");

could you remove the trailing spaces?

etienneJr:
Well, I really don't know why I put those spaces there, but anyway, I've corrected them. Thanks for the review.

I also checked that I could add type in the tiles by using this new OsmType() function (using [this lua script](https://codeberg.org/cartes/serveur/src/commit/d898d0d5c11886fd9ed64ad61c170e38e3e40211/tilemaker/resources/process-openmaptiles.lua#L893)), and it worked as expected

image

Why did the cheks failed ? Do I have to correct something else ?

Thanks !

PR #838, opened 2025-07-21
Allow `combine_below` on each layer

etienneJr:
### Context
The fact that `combine_below` is defined in the `settings` part of `config.json` is not very flexible, and not consistent with `combine_points` and `combine_polygons_below` properties, defined in each layer.

### Solution
Following #837 I tried to copy the logic of `combinePoints` and `combinePolygonsBelow` layer properties to create a new property `combineLinesBelow` that will store the value read in each layer definition in `config.json`.

But it is not working 😥​

`tile_worker.cpp` is able to read the old `ld.combinePolygonsBelow` property but is not able to read the new `ld.combineLinesBelow` property. `shared_data.cpp` and `shared_data.h` seems fine to me, I just copied the logic of `combinePolygonsBelow`, I really can't see what I've missed. `ProcessObjects()` works fine if I force a value for its new boolean property `combineLines`, but in `ProcessLayer()` the new Layer property `ld.combineLinesBelow` value is always 0.

Could you help me correct this? Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,12 @@ Bug reports, suggestions and (especially!) pull requests are very welcome on the
 
 Formatting: braces and indents as shown, hard tabs (4sp). (Yes, I know.) Please be conservative about adding dependencies or increasing the memory requirement.
 
+To safely rebuild the project after making changes to the code, you need to run :
+
+    make clean
+    make
+    sudo make install
+
 ## Copyright
 
 tilemaker is maintained by Richard Fairhurst and supported by [many contributors](https://github.com/systemed/tilemaker/graphs/contributors). We particularly celebrate the invaluable contributions of Wouter van Kleunen, who passed away in 2022.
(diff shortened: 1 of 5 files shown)


etienneJr:
> (Silly question - have you tried doing a make clean before building? The build doesn't always pick up changes to .h files otherwise.)

Oh no, it's not a silly question at all!! It's my first C++ project, I didn't know I had to start with a `make clean`! It resolved all the errors I had. I feel rather foolish, I wasted so much time trying to understand my mistake in the code 😵​

> only suggestion I'd make is renaming `combine_below` to `combine_lines_below` in the layer JSON. I think we can then deprecate `combine_below` as a global option.

I wanted to keep `combine_below` in the global settings for retrocompatibility. In the layers settings, I used `combine_lines_below`. If `combine_lines_below` is absent in the layer settings, we use the global setting `combine_below`. Is it fine for you ?

I will finish the corrections, tests, and documentation tomorrow.

etienneJr:
Hi Richard,
I pushed (force) a single clean commit with the feature and its documentation. You can see the expected result in the comment above. It's fine on my side, I remove the WIP.

PR #866, opened 2025-12-19
Add new function to allow ID modification

etienneJr:
Follwing #865 and assisted by Richard's advice, I propose this PR, which creates a new `ModifyId(newId)` function that allows any int value to be assigned to the current object in the tile.

On this stupid example, I put `123456` to all the `route` objects.
image


etienneJr:
In https://codeberg.org/cartes/serveur/pulls/112 I have included some videos showing what this PR makes possible when ID is overridden : clicking on a label directly opens the corresponding relation (and not the `waterway` way or the `admin_centre` node as before)

etienneJr:
@systemed would you have time to review and merge this PR? I will soon need it for cartes.app. Thanks!

systemed:
Done - thank you!

PR #874, opened 2026-01-22
* src/tile_worker.cpp: ignore points too far out of the valid tile.

prgm767:
This makes a pmtiles of the entire world around 10% smaller.

Not sure if the 256 buffer is needed.

### DIFF
--- a/src/tile_worker.cpp
+++ b/src/tile_worker.cpp
@@ -293,16 +293,32 @@ void ProcessObjects(
 			// so that we can write a multipoint instead of many point features
 
 			std::vector<std::pair<int, int>> multipoint;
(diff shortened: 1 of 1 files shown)


systemed:
Sorry to take a while getting back to this! Having done a bit of reading around I agree that there isn't any benefit in using the large buffer for point features. I suspect we could actually do without the 256 buffer entirely, but let's leave that in for now and maybe revisit later. Thanks for the patch.

prgm767:
thank you for merging, BTW I experimented with doing the same with polygons and linestrings, but the savings were so small, just a couple of megabytes, that it did not seem worth it.

PR #885, opened 2026-04-05
On Ubuntu and Debian, suggest using lua 5.3 rather than 5.1

SomeoneElseOSM:
lua 5.3 should be available for any recent Ubuntu or Debian. Reasons for preferring 5.3 over 5.1 include the availability of UTF8 libraries.  If 5.1 is explicitly installed alongside 5.3 it's possible to end up with different versions with different shared library locations, which can confuse.

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -17,7 +17,7 @@ Then:
 
 Start with:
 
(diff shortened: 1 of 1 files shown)


SomeoneElseOSM:
Odd that the CI failed when I changed one .md file...

systemed:
Thanks!

PR #886, opened 2026-05-11
Fix Build Issues

Symmetricity:
This PR is AI generated.

## Summary

This restores the ability to build from source across multiple operating systems by resolving several outstanding issues.

- replace an MSVC-only `uint` build failure in `OutputObject`
- keep linking `boost_system` in the Makefile when the library exists, but omit it when Boost no longer ships it
- link `Boost::throw_exception` from CMake when that imported target is available
- include `<cstdint>` from `helpers.h`, which directly uses fixed-width integer types

Fixes #880.
Fixes #882.

## Background

Recent builds fail:

- Windows CMake fails in `include/output_object.h` because `uint` is no longer available after the Boost 1.89 compatibility changes
- macOS Makefile jobs fail because Homebrew Boost 1.90 no longer provides `libboost_system`

This follows the Boost 1.89+ compatibility work in #871 and the Boost.System removal reports in #862 and #880.

## Testing

- `cmake -S . -B build-ci-fix -DCMAKE_BUILD_TYPE=RelWithDebInfo`
- `cmake --build build-ci-fix --target tilemaker -j2`
- `make -j4`
- `make test`
- Windows CMake/MSVC/vcpkg build with:
  - `cmake -DTILEMAKER_BUILD_STATIC=ON -DVCPKG_TARGET_TRIPLET=x64-windows-static-md -DCMAKE_TOOLCHAIN_FILE=C:\vcpkg\scripts\buildsystems\vcpkg.cmake ..`
  - `cmake --build . --config RelWithDebInfo`

Issue #882, opened 2026-03-19
error: 'uint' has not been declared


ryandesign:
We received [a bug report](https://trac.macports.org/ticket/73678) at MacPorts from a user who cannot build tilemaker 3.1.0 on Mac OS X 10.6:

```
[ 57%] Building CXX object CMakeFiles/tilemaker.dir/src/output_object.cpp.o
/opt/local/bin/g++-mp-14 -DBOOST_ALL_NO_LIB -DBOOST_FILESYSTEM_DYN_LINK -DBOOST_PROGRAM_OPTIONS_DYN_LINK -DBOOST_SYSTEM_DYN_LINK -I/opt/local/var/macports/build/tilemaker-c46fb254/work/tilemaker-3.1.0/include -I/opt/local/var/macports/build/tilemaker-c46fb254/work/build -isystem /opt/local/include -isystem /opt/local/libexec/boost/1.76/include -pipe -I/opt/local/libexec/boost/1.76/include -Os -DNDEBUG -I/opt/local/libexec/boost/1.76/include -I/opt/local/include -D_GLIBCXX_USE_CXX11_ABI=0 -std=gnu++17 -arch ppc -mmacosx-version-min=10.6 -MD -MT CMakeFiles/tilemaker.dir/src/output_object.cpp.o -MF CMakeFiles/tilemaker.dir/src/output_object.cpp.o.d -o CMakeFiles/tilemaker.dir/src/output_object.cpp.o -c /opt/local/var/macports/build/tilemaker-c46fb254/work/tilemaker-3.1.0/src/output_object.cpp
In file included from /opt/local/var/macports/build/tilemaker-c46fb254/work/tilemaker-3.1.0/src/output_object.cpp:5:
/opt/local/var/macports/build/tilemaker-c46fb254/work/tilemaker-3.1.0/include/output_object.h:30:17: error: 'uint' has not been declared
   30 |                 uint mz
      |                 ^~~~
make[2]: *** [CMakeFiles/tilemaker.dir/src/output_object.cpp.o] Error 1
```

Although I admit that Mac OS X 10.6 is old, I suspect this can be made to work.

I checked both Mac OS X 10.6 and macOS 12, and `uint` is defined in the `<sys/types.h>` header:

```c
#if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
#include <sys/_types/_u_char.h>
#include <sys/_types/_u_short.h>
#include <sys/_types/_u_int.h>
[...]

Issue #864, opened 2025-11-30
tilemaker generates degenerate polygons not present in source


breunigs:
Given this way in version 1 as input: https://www.openstreetmap.org/way/1422845261/history (a thin, L-shaped, valid polygon)

as source files for convenience:
- [bug.osm.gz](https://github.com/user-attachments/files/23839453/bug.osm.gz)
- [bug.osm.pbf.gz](https://github.com/user-attachments/files/23839454/bug.osm.pbf.gz)

and this very minimal configuration:

config.json
```json
{
	"layers": {
		"building": {
			"minzoom": 14,
			"maxzoom": 14
		}
	},
	"settings": {
		"minzoom": 14,
		"maxzoom": 14,
		"basezoom": 14,
		"include_ids": true,
		"name": "debug",
		"version": "3.0",
		"description": "",
		"compress": "gzip"
	}
}
```

process.lua
```lua
function way_function()
  Layer("building", true)
end
```

and generating tiles from it:

```bash
./tilemaker \
  --input bug.osm.pbf \
  --output bug.mbtiles \
  --config ./config.json \
  --process ./process.lua \
  --bbox 10.0534129,53.6191617,10.0544429,53.619693 # same as from bug.osm

sqlite3 \
  bug.mbtiles \
  'select quote(tile_data) from tiles where zoom_level=14 and tile_column=8649 and tile_row=11094' \
  | cut -d\' -f2 \
  | xxd -r -p > bug.mvt

vt2geojson -z 14 -y 5289 -x 8649 bug.mvt
```

yields a degenerate polygon:

```json
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [
              10.053927898406982,
              53.61927620546507
            ],
            [
              10.053858160972595,
              53.61927620546507
            ],
            [
              10.053799152374268,
              53.61927620546507
            ],
            [
              10.053927898406982,
              53.61927620546507
            ]
          ]
        ]
      },
      "properties": {},
      "id": 1422845261
    }
  ]
}
```

The [spec technically allows degenerate polygons](https://github.com/mapbox/vector-tile-spec/blob/master/2.1/README.md#4344-polygon-geometry-type), but discourages them:

> A linear ring SHOULD NOT have an area calculated by the surveyor's formula equal to zero, as this would signify a ring with anomalous geometric points.

It also includes wording like

[...]

Issue #862, opened 2025-11-23
Boost 1.89 compatibility issue w/tilemaker (macOS)


johncarmack1984:
In [Boost 1.89](https://www.boost.org/releases/1.89.0/) (released 8/6/2025), there were several changes to the Geometry library used by tilemaker. 

This included [removing static variables from header files](https://github.com/boostorg/geometry/pull/1405), which has caused some issues in tilemaker when using `#include <boost/sort/sort.hpp>` (ie in `tilemaker/include/tile_data.h`, see [this failed workflow's](https://github.com/systemed/tilemaker/actions/runs/19361803292/job/55395312337) error messages). 

This can be circumvented by using Boost 1.88. 

However, homebrew install does not support this; one must either build Boost from source OR install it using a custom tap with the archived formula (which is both the [recommended solution](https://docs.brew.sh/Versions) in their docs and [considered counter to homebrew's philosophy](https://github.com/orgs/Homebrew/discussions/5454) by at least one maintainer). 

This increases the complexity of building tilemaker on macOS significantly. 

It would seem a refactor of tilemaker's `boost/sort` headers is needed to accomodate the new structure of `boost/geometry`, but accomplishing this feat proved beyond what my (admittedly novice) C++ skills could solve in a weekend.

Any thoughts or tips one might give to make this an easily-solved problem would be welcomed, and if I can solve it, will be happy to PR, but lacking that, here's an issue to note that the problem has been identified & diagnosed. :)

AlanIWBFT:
Probably more of a boost problem: https://github.com/boostorg/geometry/issues/1445

AlanIWBFT:
Meanwhile, you can workaround the issue by:

1. Move the definition of `finalizeObjects` from `tile_data.h` into `tile_data.cpp`, before `TileDataSource::finalize`. Then move the `sort.hpp` include into the cpp file as well.
2. Move the `sort.hpp` include in `osm_store.cpp` to be the very first include.
3. Not really related, but `system` has been removed in boost 1.89. Remove `system` and `Boost::system` from `CMakeList.txt`

Now the project should compile.

PR #887, opened 2026-05-13
Improve CI Workflow

Symmetricity:
# Improve CI Workflow

This PR was AI generated.

## Summary

This updates the GitHub Actions CI workflow so it is faster, less repetitive,
and checks more than whether tilemaker compiles.

- update GitHub Actions to current major versions
- add explicit workflow permissions and concurrency cancellation
- cache the Geofabrik Liechtenstein fixture using freshness metadata, validate
  it against Geofabrik's published `.md5`, then share it with build jobs as an
  artifact
- add a fast static-validation job before fixture download and build-heavy jobs
- skip build-heavy jobs when a change cannot affect runtime behavior
- use runner-image-scoped vcpkg binary caches instead of caching the installed
  vcpkg tree directly
- add Docker Buildx cache and split PR image builds from `master` publishes
- generate MBTiles and PMTiles outputs from each CI build path
- fail immediately if a tilemaker invocation fails or does not create the
  expected output archive
- run tile generation twice per build path and verify repeatability
- upload generated tile archives for inspection
- verify PMTiles archive structure with `pmtiles verify`
- compare generated tile contents semantically, not just byte-for-byte
- normalize MVT geometry encodings during semantic comparison, including
  point ordering and polygon ring rotation while preserving polygon ring
  grouping
- report progress and ETA during expensive semantic tile comparisons
- reuse per-archive semantic tile caches when comparing generated outputs so
  each archive is decoded once even when it is compared against several others
- keep the published Docker action output out of generated-tile validation
  because it is built from the published `master` image, not the PR build

## Rationale

The old workflow downloaded the same PBF independently in several jobs and only
verified that the build commands completed. That made CI slower than necessary
and left important runtime behavior unchecked.

The shared fixture job makes every build use the same input file. It keys the
Actions cache from Geofabrik response metadata (`URL`, effective URL, `ETag`,
`Last-Modified`, and `Content-Length`) so repeat runs can reuse the fixture
without pinning CI to an old file forever. When Geofabrik republishes the PBF,
the metadata-derived key changes and CI downloads the new fixture. The restored
or downloaded fixture is validated against Geofabrik's published `.md5` before
it is uploaded for build jobs. If Geofabrik is unavailable or the checksum does
not match, the workflow reports that tile generation was not verified rather
than making the failure look like a code regression.

The vcpkg cache now stores vcpkg binary packages and includes the GitHub runner
image identity in the key. That avoids blindly restoring an installed tree from
a different runner image while still allowing warm dependency restores on stable
runners. Docker builds also use the GitHub Actions build cache.

The generated tile verification exists because tilemaker's most important output
is the generated vector-tile archive. A build can compile cleanly while still
producing invalid PMTiles, non-repeatable tile content, or different content
between build paths. CI should catch those cases.

Each tilemaker invocation is now checked as soon as it returns. This avoids
multi-command CI steps masking a failed archive generation when a later command
succeeds, and it makes the failing output visible in the generation job rather
than later as a missing artifact or verifier failure.

The static-validation job gives the workflow a cheaper failure path for changes
to CI, profiles, and scripts. It validates GitHub workflow files with
`actionlint`, shell scripts with `shellcheck`, tracked JSON files with Python,
Lua profiles with `luac`, and Python CI helper scripts with `py_compile` before
starting the fixture download, Docker builds, or C++ build matrix.

## Output verification

The new verifier checks two levels of output equality:

- raw decompressed tile bytes, which catches byte-for-byte differences after
  removing archive compression noise
- semantic MVT content, which canonicalizes layer order, feature order, and tag
  order before comparing. It also decodes geometry command streams so equivalent
  point ordering and polygon ring rotation do not fail as content changes, while
  preserving polygon outer-ring and inner-ring grouping.

Raw byte differences with identical semantic content are reported as notices.
Semantic tile-content differences fail the job and include the first differing
tile. When the layer feature counts differ, the error also reports the affected
layer and counts. The reported tile is taken from the current comparison, not
from unrelated semantic tiles already present in the per-archive cache.

Semantic comparisons populate a per-archive tile-content cache. That keeps
MBTiles and PMTiles on the same comparison path and avoids repeatedly decoding
the same archive when a verifier run checks repeat outputs and then cross-runner
outputs.

PMTiles archives are also checked with `pmtiles verify` before their tile content
is compared. The PMTiles verifier output is included in the GitHub annotation so
the failure is visible in the web UI without digging through raw logs.

The GitHub Action job uses the published
`ghcr.io/systemed/tilemaker:master` Docker image from `action.yml`, not the
PR-built CMake or Makefile binaries. It is therefore kept out of generated-tile
validation. On `master`, it runs after `docker-publish` as a smoke test for the
published Action image.

## Current findings

The verifier was run locally against artifacts downloaded from fork CI runs.
Those tests showed that the new check is catching real output issues, not just
formatting or compression differences:

- Windows PMTiles archives failed structural verification with a header length
  mismatch reported by `pmtiles verify`.
- Repeat runs produced semantic tile-content differences, including examples
  where the same tile/layer had different feature counts, such as `poi` having
  `102` features in one run and `101` in the repeat run.
- Many byte-level differences were only ordering differences, which is why the
  verifier now separates raw-byte mismatches from semantic MVT mismatches.
- Some raw MVT geometry differences were equivalent encodings of the same
  feature geometry; the verifier now normalizes those before calculating
  semantic hashes.
- One fork CI run uploaded an incomplete generated-tile artifact after a build
  job produced only three of the four expected archives. The workflow now fails
  at the tile generation step if an expected archive is missing or empty.
- A later fork CI run with the fail-fast wrapper caught the same class of
  problem at source: Windows CMake exited with `-1073741819` while writing the
  stored `liechtenstein.mbtiles` output, instead of uploading a partial artifact.
- A larger local Austria PMTiles repeat test produced `47,742` addressed tiles,
  with `43,124` raw-byte-different tiles between warmed repeat runs. The
  semantic verifier confirmed that semantic content matched, and later verifier
  profiling showed that reusing per-archive semantic caches avoids repeated
  archive decoding when three or more outputs are compared.

These findings are not fixed by this PR. This PR adds CI coverage that exposes
them consistently.

## Related issues and PRs

This PR does not include the build-fix changes from #886. If `master` still has
those build failures when this PR is tested, this PR should be applied after
#886 or rebased once equivalent fixes land.

Directly related:

- #886 fixes current source build failures; this PR intentionally keeps those
  code/build-system fixes separate.
- #880 reports modern Boost/Homebrew `boost_system` linker failures; covered by
  #886, not this PR.
- #882 reports `uint` portability failures; covered by #886, not this PR.
- #862 and #871 are the prior Boost 1.89 compatibility context.

This PR may improve future detection and debugging for output-related issues,
but does not claim to fix them:

- #653 PMTiles root/leaf index size issue
- #697 invalid polygons after clipping
- #739 ocean tiles incomplete
- #847 woods and water disappearing at z14
- #864 degenerate polygons not present in source

No directly relevant GitHub Discussions were found for deterministic generated
tile output checks.

## Testing

- `actionlint .github/workflows/ci.yml`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/ci.yml"); puts "yaml ok"'`
- `git ls-files -z -- '*.sh' | xargs -0 -r shellcheck`
- `git ls-files -z -- 'resources/*.lua' | xargs -0 -r luac5.4 -p`
- tracked JSON files parsed with Python `json.load`
- `git ls-files -z -- '.github/scripts/*.py' | xargs -0 -r python3 -m py_compile`
- `python3 -B -c 'import py_compile; py_compile.compile(".github/scripts/verify-generated-tiles.py", cfile="/tmp/verify-generated-tiles.pyc", doraise=True)'`
- `git diff --check`
- synthetic verifier check for stale cached semantic tiles in error reporting
- local verifier run against downloaded fork CI tile-output artifacts
- local verifier run against two warmed Austria PMTiles outputs; raw tile bytes
  differed, semantic content matched
- local verifier timing against downloaded fork CI artifacts showed the cached
  comparison path preserving the same failure result while reducing wall time

The clean branch currently contains only CI-related files:

- `.github/workflows/ci.yml`
- `.github/scripts/verify-generated-tiles.py`
- `.gitignore`

### DIFF
--- a/.github/scripts/verify-generated-tiles.py
+++ b/.github/scripts/verify-generated-tiles.py
@@ -694,10 +694,12 @@ def tile_label(tile):
     return f"z{z}/{x}/{y}"
 
 
-def tile_difference(left, right, semantic):
+def tile_difference(left, right, semantic, tiles=None):
     left_tiles = left.semantic_tiles if semantic else left.raw_tiles
     right_tiles = right.semantic_tiles if semantic else right.raw_tiles
-    for tile in sorted(set(left_tiles) | set(right_tiles)):
+    if tiles is None:
+        tiles = sorted(set(left_tiles) | set(right_tiles))
+    for tile in tiles:
         if tile not in left_tiles:
             return f"{tile_label(tile)} is missing from output"
         if tile not in right_tiles:
@@ -782,7 +784,7 @@ def compare_content(path, content, other_path, other_content, relation, semantic
             error(
                 path,
                 f"semantic content differs from {relation} {archive_label(other_path)}; "
-                f"{tile_difference(content, other_content, True)}",
+                f"{tile_difference(content, other_content, True, [tile])}",
             )
             return
         if index % PROGRESS_INTERVAL == 0 or index == len(differing_tiles):
--- a/.github/scripts/verify-generated-tiles.py
+++ b/.github/scripts/verify-generated-tiles.py
@@ -127,6 +127,10 @@ def archive_label(path):
     return f"{build_label(path)} {path.name}"
 
 
+def should_validate_archive(path):
+    return path.parent.name != "tile-outputs-github-action"
+
+
 def decompress_payload(data):
     for wbits in (16 + zlib.MAX_WBITS, zlib.MAX_WBITS):
         try:
@@ -973,8 +977,8 @@ def check_cross_runner(contents, suffix, semantic_decoder, failure_message, sema
 
 def main():
     root = pathlib.Path(sys.argv[1])
-    mbtiles_paths = sorted(root.glob("**/*.mbtiles"))
-    pmtiles_paths = sorted(root.glob("**/*.pmtiles"))
+    mbtiles_paths = sorted(filter(should_validate_archive, root.glob("**/*.mbtiles")))
+    pmtiles_paths = sorted(filter(should_validate_archive, root.glob("**/*.pmtiles")))
 
     print("Archive SHA-256")
     mbtiles = fingerprint_archives(
--- a/.github/scripts/verify-generated-tiles.py
+++ b/.github/scripts/verify-generated-tiles.py
@@ -1,18 +1,29 @@
 #!/usr/bin/env python3
 
+import concurrent.futures
 import gzip
 import hashlib
 import json
+import os
 import pathlib
 import sqlite3
 import struct
 import subprocess
 import sys
+import time
 import zlib
 
 
 COMPRESSION_NONE = 1
 COMPRESSION_GZIP = 2
+PROGRESS_INTERVAL = 1024
+SEMANTIC_COMPARE_CHUNK_SIZE = 128
+
+
+worker_left_archive = None
+worker_right_archive = None
+worker_left_compression = None
+worker_right_compression = None
 
 
 status = 0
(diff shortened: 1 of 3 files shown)


systemed:
Neither CI nor Python are things I speak fluently (or at all, really), so this is a bit above my paygrade. But I like the idea of checking the generated tiles, the CI isn't in a great state currently, and I can't see any obvious malware in the script :), so let's give this a go and see how it turns out. Thank you!

Issue #785, opened 2024-11-24
Polylabel algorithm operation


pavel-avilov:
Hi, could you please explain how does the polylabel algorithm work in the LayerAsCentroid/Centroid methods.

As I understand, the polylabel algorithm should find the point that is the most distant from any edge. I've noticed that some of the house numbers are located a little strangely. I've used the LayerAsCentroid method (just like in your process-openmaptiles.lua) to save the house number info.

Examples of strange housenumber placement with default styles:
<img width="361" alt="hn1" src="https://github.com/user-attachments/assets/b711881b-dfae-4d49-b8ba-8d68bcb3ec67">
<img width="361" alt="hn2" src="https://github.com/user-attachments/assets/ba083724-61db-4c9b-8793-868a47af68ca">

For example, on the last screenshot the house number for [way](https://www.openstreetmap.org/way/501460502) is located at the following coordinates:
lat: 51.536720348065
lon: -0.1600334265625
<img width="316" alt="hn3" src="https://github.com/user-attachments/assets/49554f0c-4baf-4010-8f1f-e988a20c8a7e">
In OSM it looks like this:
<img width="452" alt="osm1" src="https://github.com/user-attachments/assets/02fbee79-446c-4d26-b8cf-b0fade117748">
I decided to double-check the work of polylabel on this house using [Python-Polylabel](https://github.com/Twista/python-polylabel)

I ran the following:
```
polylabel([[51.5372152, -0.1604443], [51.5371332, -0.1606558], [51.5365747, -0.1600228], [51.5365893, -0.1599317], [51.5366517, -0.1597992], [51.5372152, -0.1604443]])
```
<details>
<summary>The code as I checked:</summary>

```
from OSMPythonTools.overpass import Overpass
from polylabel import polylabel

overpass = Overpass()
way_id = '501460502'
query = f'way({way_id}); out geom;'
result = overpass.query(query)
all_coord = []

# Get way coords
for way in result.ways():
    print(f'Way ID: {way.id()}')
    print('Coordinates:')
    for coord in way.nodes():
        print(f'Latitude: {coord.lat()}, Longitude: {coord.lon()}')
        all_coord.append([coord.lat(), coord.lon()])

print(f"polylabel({all_coord}) >")
print(polylabel([all_coord]))
```
</details>

and then I got
lat: 51.53688964928815
lon: -0.16022586209752077
In OSM it looks like this:
[...]

PR #890, opened 2026-05-13
Handle open way centroids as lines

Symmetricity:
This PR is AI generated.

## Handle open way centroids as lines

`LayerAsCentroid()` defaults to the `polylabel` algorithm for 2D geometries.
That is appropriate for polygonal areas, but open ways are linestrings. The
current code assigns every way into a polygon before calculating the
representative point, so open ways can be passed through polygon/polylabel
logic even though they are not closed areas.

This change keeps the existing polygon path for closed ways, but handles open
ways from the cached linestring directly:

- empty open ways keep the existing error path by throwing
  `geom::centroid_exception`
- single-point open ways use that point
- multi-point open ways use Boost.Geometry's linestring centroid

Closed ways and relations still use the existing polygon/polylabel behavior.
The stored output is still a point, so this changes how the point is calculated
for open ways without changing output layers, attributes, or geometry storage.

This also avoids a temporary polygon conversion for open ways, but the intent
is correctness rather than a performance optimization.

## Reproduce

Generate OpenMapTiles-compatible output for an extract that contains open ways
which are emitted through `LayerAsCentroid()`, for example named POI-like ways
handled by `WritePOI()` or the catch-all `poi_detail` path in
`process-openmaptiles.lua`.

Before this change, those open ways are assigned into a polygon and labeled via
polygon logic. That can produce representative points that depend on polygon
handling for invalid/open rings.

After this change, open ways use line geometry for their representative point,
while closed ways continue to use polygon labeling.

In the generated-tile CI investigation, the focused open-way centroid patch
changed only `poi` geometry in the local Liechtenstein debug output. Tile
counts, feature counts, attributes, and polygon layers were unchanged in that
run.

## Testing

- `cmake --build build -j2`

## Related Issues And PRs

I did not find an existing upstream issue or PR that directly reports open ways
being labeled through polygon/polylabel handling.

Related centroid/polylabel context:

- #785 discusses `polylabel` behavior and precision for `LayerAsCentroid()` /
  `Centroid()`.
- #392 discusses point-on-surface style label placement for polygons.
- #641 fixed other `LayerAsCentroid()` behavior around Lua interop and relation
  members.

### DIFF
--- a/src/osm_lua_processing.cpp
+++ b/src/osm_lua_processing.cpp
@@ -862,14 +862,25 @@ Point OsmLuaProcessing::calculateCentroid(CentroidAlgorithm algorithm) {
 		}
 		return Point(centroid.x()*10000000.0, centroid.y()*10000000.0);
 	} else if (isWay) {
-		Polygon p;
-		geom::assign_points(p, linestringCached());
-
-		if (algorithm == CentroidAlgorithm::Polylabel) {
-			// CONSIDER: pick precision intelligently
-			centroid = mapbox::polylabel(p);
+		if (!isClosed) {
+			const Linestring &ls = linestringCached();
+			if (ls.empty())
+				throw geom::centroid_exception();
+			if (ls.size()==1) {
+				centroid = ls.front();
+			} else {
+				geom::centroid(ls, centroid);
+			}
 		} else {
-			geom::centroid(p, centroid);
+			Polygon p;
+			geom::assign_points(p, linestringCached());
+
+			if (algorithm == CentroidAlgorithm::Polylabel) {
+				// CONSIDER: pick precision intelligently
+				centroid = mapbox::polylabel(p);
+			} else {
+				geom::centroid(p, centroid);
+			}
 		}
 		return Point(centroid.x()*10000000.0, centroid.y()*10000000.0);
 	} else {


Symmetricity:
I also checked the change visually against the OSM source geometry from the same
Liechtenstein PBF fixture.

Legend in the screenshots:

- red point: upstream `master`
- green point: this PR
- blue line: the OSM source way from the PBF
- OSM raster tiles: background context only

For the visual check I used a temporary debug process file with stable POI
ordering and a `debug_osm_id` attribute, so the comparison isolates this C++
change from Lua table-order noise. With that setup, only `poi` point placement
changed; global layer counts stayed the same, and no polygon/line/building/
housenumber layers changed.

The largest example is `Finnenbahn`, where upstream places some labels away
from the source way after treating an open way as polygon input. This PR places
those labels on the corresponding OSM line geometry:

Finnenbahn open-way centroid comparison

Measured against the source way in the PBF:

- OSM way `424087004`: upstream 26.1 m from the source line, PR 0.3 m
- OSM way `424087005`: upstream 5.1 m from the source line, PR 0.1 m
- OSM way `424087003`: upstream 0.3 m from the source line, PR 0.0 m

Two smaller examples:

Climbing open-way centroid comparison

Bollard open-way centroid comparison

Overview of the checked examples:

Open-way centroid comparison overview

systemed:
I hadn't considered the use case of centroids on linestrings but that's a good spot - thank you.

Issue #121, opened 2018-10-15
How to configure fields?


wboykinm:
I'm interested in pulling building height data into MVTs generated with tilemaker. Using the "default" config file, I edited [this line](https://github.com/systemed/tilemaker/blob/master/config.json#L21) to read

`{ "id": "building", "description": "building", "fields": { "height": "Integer" }}`

Tilemaker then runs smoothly and produces outputs [that render in Mapbox GL when given static properties](https://github.com/wboykinm/bird/blob/master/lib/styles.js#L76):

<img width="620" alt="screenshot 2018-10-15 10 29 13" src="https://user-images.githubusercontent.com/735463/46957284-40aaa680-d065-11e8-974a-a010cb60222b.png">

Unfortunately, when I try to [use the height info for dynamic styling](https://github.com/wboykinm/bird/blob/master/lib/styles.js#L77-L85), I get this error:

`Expected value to be of type number, but found null instead.`

It appears that my adjustment to the config file above is producing `NULL`s. What is the proper way to pass usable building height data (or indeed any data from osm properties) through to generated MVTs?

systemed:
Oh wow, that looks nice!

The bit of tilemaker that maps OSM keys into MVT fields is the `config.lua` file. (The `vector_layers` part of the JSON config is a slightly confusing Mapbox addition which documents what the client will find in the tiles, but it doesn't actually change anything. tilemaker can now generate that automatically.)

So you'd need to make `config.lua` take the OSM data and write it into the vector tiles. You should be able to do this by editing https://github.com/systemed/tilemaker/blob/master/process.lua#L43, and adding a couple of new lines:

    local metres = to_number(way:Find("height")) or 5
    way:AttributeNumeric("height", metres)

That takes the height tag from OSM; converts it to a number; falls back to a default height of 5 if there's no height tag; and writes it to the vector tile as the "height" attribute.

There's doubtless lots more you could do in terms of parsing the weird and wonderful values that are no doubt in the height tag, but that should do for a start. Lua's really well-suited to doing this sort of OSM tag-mangling (it's used for the same purpose in OSRM and osm2pgsql).

wboykinm:
@systemed That's a marvelous explanation, and makes this super-extensible. Thanks so much! 

I am however running into what appears to be a coalesce failure, and my troglodytic searching around the lua docs is not turning up an answer. 

Given:

```
if building~="" then
  local metres = (to_number(way:Find("height"))) or 5
[...]

PR #891, opened 2026-05-14
Parse decimal building heights

Symmetricity:
This PR is AI generated.

## Parse decimal building heights

The OpenMapTiles profile parses `height`, `min_height`, `building:levels`, and
`building:min_level` with `tonumber(value, 10)`.

That works for integer tags, but newer Lua versions reject decimal strings when
an explicit base is supplied. Valid OSM values such as `height=18.5` and
`building:levels=3.5` therefore parse as `nil` and fall back to the default
one-floor render height. LuaJIT accepts the same values with the base argument,
so the same input can produce different `render_height` values depending on the
Lua runtime used to build tilemaker.

Parse these four fields with plain `tonumber(value)` instead. That keeps
integer handling unchanged, accepts decimal OSM values, and makes the profile
consistent across Lua runtimes.

The change is intentionally scoped to the OpenMapTiles building height helper.
It does not alter layer selection, building feature counts, minzoom handling, or
the existing fallback height for buildings with no usable height/level tags.

## Examples

The examples below are from Liechtenstein CI artifacts generated from the same
input.

### Decimal `height`

Source object:

```
way 490877222
height=18.5
building=detached
```

OSM link:

```
https://www.openstreetmap.org/way/490877222
```

Pre-fix output on a newer Lua runtime:

```
render_height=3.66
```

Pre-fix output on LuaJIT:

```
render_height=18.5
```

Post-fix output:

```
render_height=18.5
```

Decimal height comparison

### Decimal `building:levels`

Source object:

```
way 165295920
building:levels=3.5
building=school
```

OSM link:

```
https://www.openstreetmap.org/way/165295920
```

Pre-fix output on a newer Lua runtime:

```
render_height=3.66
```

Pre-fix output on LuaJIT:

```
render_height=12.81
```

Post-fix output:

```
render_height=12.81
```

Decimal building levels comparison

These examples show both parts of the bug: decimal tags were being ignored on
some Lua runtimes, and generated tile content could differ across CI runners
for the same input.

Building height overview

## Reproduce

With a newer Lua runtime:

```
lua -e 'print(tonumber("18.5", 10)); print(tonumber("18.5"))'
```

Before this change, the first expression returns `nil`, so
`SetBuildingHeightAttributes()` falls back to `BUILDING_FLOOR_HEIGHT`.

After this change, decimal height and level values parse as numbers and the
rendered building attributes use the source tag values.

## Testing

- `luac -p resources/process-openmaptiles.lua`
- Focused Lua check that `tonumber("18.5", 10)` is rejected by the local newer
  Lua runtime while `tonumber("18.5")` is accepted
- Local generated-tile artifact comparison for decimal `height` and decimal
  `building:levels` examples

## Related Issues And PRs

Related upstream context:

- #121 includes prior maintainer guidance to parse OSM height tags with plain
  `tonumber`.
- PR #263 added the current `render_height` and `render_min_height` attributes
  for 3D building rendering.
- PR #692 is an open building-related PR that expands `building:part` handling,
  but it does not directly address this Lua-runtime decimal parsing issue.

I did not find an existing upstream issue, PR, or discussion that directly
reports decimal building heights being ignored on newer Lua runtimes.

### DIFF
--- a/resources/process-openmaptiles.lua
+++ b/resources/process-openmaptiles.lua
@@ -868,10 +868,10 @@ function GetPOIRank()
 end
 
 function SetBuildingHeightAttributes()
-	local height = tonumber(Find("height"), 10)
-	local minHeight = tonumber(Find("min_height"), 10)
-	local levels = tonumber(Find("building:levels"), 10)
-	local minLevel = tonumber(Find("building:min_level"), 10)
+	local height = tonumber(Find("height"))
+	local minHeight = tonumber(Find("min_height"))
+	local levels = tonumber(Find("building:levels"))
+	local minLevel = tonumber(Find("building:min_level"))
 
 	local renderHeight = BUILDING_FLOOR_HEIGHT
 	if height or levels then


systemed:
Thank you!

PR #901, opened 2026-05-22
Write rings in opposite order since Y axis is inverted (relative to B…

dougkpowers:
I'm rendering tiles using ESRI.  ESRI's ArcGIS JS renderer is very strict.  If Polygons are not clockwise oriented, then they will render, but they won't fill.  The reason is that when counter-clockwise, ESRI assumes the ring is an exclusion.

Tilemaker is writing almost all polygons backwards (the rings are in counter-clockwise orientation).  This is because after simplification or clipping, we call geometry::correct on the polygon.  This orients the polygon correctly according to Boost.  However, the Y value of points are inverted in vector tiles relative to a normal coordinate plane.  Boost doesn't know this, so as soon as we "correct" a polygon, the orientation is backwards.

This PR fixes the problem by doing two things:
1. We call geometry::correct on all polygons just before writing them to the vector tile.
2. We change the writeRing algorithm to traverse the the rings backwards

Doing these two things solves the problem and now the ESRI renderer is happy.

### DIFF
--- a/src/tile_worker.cpp
+++ b/src/tile_worker.cpp
@@ -192,15 +192,18 @@ bool writeRing(
 
 	bool firstPoint = true;
 	fbuilder.add_ring(points);
-	for (const Point& point : ring) {
-		pair<int, int> xy = std::make_pair(point.get<0>(), point.get<1>());
+	for (auto it = ring.rbegin(); it != ring.rend(); ++it) {
+    	const Point& point = *it;
 
-		if (firstPoint || xy != lastXy) {
-			firstPoint = false;
-			lastXy = xy;
-			fbuilder.set_point(xy.first, xy.second);
-		}
-	}
+    	pair<int, int> xy =
+    		std::make_pair(point.get<0>(), point.get<1>());
+
+    	if (firstPoint || xy != lastXy) {
+    		firstPoint = false;
+    		lastXy = xy;
+    		fbuilder.set_point(xy.first, xy.second);
+    	}
+    }
 
 	return true;
 }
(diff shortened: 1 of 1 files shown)


systemed:
Good spot on the winding order - thank you!

I am a little nervous about the extra `geom::correct` - it seems like it has the potential to introduce slowdowns on large datasets. That said, a quick test on a UK extract didn't show any discernible difference. I might try it with a larger extract but would be interested in any thoughts.

dougkpowers:
Two thoughts
1) It only appeared to be an issue for me if the geometry was being simplified (or clipped).  Otherwise, the winding order was written correctly.  I haven't studied the entire pipeline in tilemaker closely enough to understand why, but it could be that the winding order is only messed up when we call geom::correct().  If this is the case, you could simply follow geom::correct() with a 2nd transformation to reverse the order of points in the rings.  In this case, you'd only do the extra work when absolutely necessary.
2) I'm processing all of North America and measured no meaningful difference in processing time.  But my measurement was simply anecdotal.  I didn't use a stopwatch!

geoneutrino:
during my tests: complete planet calculation on my machine including Shapefiles and writing mbtiles file 3h 7m and 3h 12m last week and with this today 3h 15m with the same planet dump. 
Did some minimal changes in process.lua in the meantime but they should not affect performance

systemed:
On Europe I get 1:24:44 with this applied, 1:22:23 without. The mbtiles is very slightly larger with this patch, which suggests we're retaining more geometries (which is a good thing!). So I'll keep an eye on it but I think this is good to go - thank you!

PR #902, opened 2026-05-27
Speed up runtime by reducing allocation churn in tile generation

Symmetricity:
This PR is AI generated.

## Summary

In a warmed Austria OpenMapTiles tile generation run, this reduced wall time by
about 6.4% and user CPU by about 9-10%. In heaptrack runs on a smaller fixture,
allocation calls fell by about 47% and temporary allocations by about 74%.

This PR reduces allocation churn and unnecessary geometry work in several hot
paths used while generating tiles.

The changes are grouped around the same goal: keep existing output behavior,
but avoid repeated temporary allocations, avoid avoidable geometry copies, and
skip cheap-to-detect no-op work before it reaches heavier geometry routines.

The largest measured effect is in tile generation with the OpenMapTiles
profile: fewer geometry/vector reallocations, fewer decoded sorted-node chunks,
and less work in clipping/indexing paths.

### Node and way lookup hot paths

- Cache a small number of recent sorted-node chunks per worker instead of only
  one chunk.
- Avoid copying geometry while indexing objects.
- Fill way geometry buffers directly instead of building temporary output and
  copying it afterward.
- Reserve OSMStore and way geometry buffers before filling them.

These changes target repeated lookups and geometry population during PBF-backed
tile generation.

### Geometry construction and scaling

- Move freshly built or corrected rings into their destination containers where
  the source is no longer needed.
- Pre-size populated multipolygons and closed-way polygon rings.
- Reserve split-linestring and Visvalingam output storage.
- Reuse projection/scaling output buffers for polygon area and scaled geometry
  calculations.

These changes reduce short-lived `std::vector` allocation/reallocation in
geometry preparation paths.

### Clipping and intersection paths

- Reuse the scratch ring used by the fast polygon clipper.
- Skip fast clipping for bounded linestrings that are already inside the clip
  box.
- Use line segments directly for line tile-intersection checks.
- Reuse dissolve intersection output storage.
- Move uncached multipolygons into the clipping path instead of copying them
  when the current function owns the geometry.
- Skip Sutherland-Hodgman edge passes when no point needs clipping against that
  edge.

These changes keep the same clipping model, but reduce repeated temporary
storage and avoid unnecessary pass-through clipping work.

## Performance

I compared current upstream master with this performance stack applied directly
on top of upstream master, without other unmerged PRs.

Branches/binaries:

```text
upstream: b437d7c
performance stack: ba8db48
```

Runtime fixture:

```text
fixture: Austria Geofabrik extract
profile: OpenMapTiles profile
output: PMTiles
store: no --store
threads: 8
warmup: PBF plus coastline/landcover sidecar files
```

Three alternating warmed runs, upstream first:

```text
wall time:   -2.19s  (-6.42%)
user CPU:    -19.34s (-9.28%)
system CPU:  +2.09s  (+6.00%)
RSS:         -47.6 MiB (-1.07%)
```

Three alternating warmed runs, performance stack first:

```text
wall time:   -2.17s  (-6.38%)
user CPU:    -20.05s (-9.75%)
system CPU:  +2.21s  (+5.93%)
RSS:         +47.4 MiB (+1.08%)
```

The wall-time and user-CPU improvement reproduced in both orders. RSS moved in
opposite directions depending on run order, so I would treat native RSS as
neutral rather than claiming a memory saving.

Perf counters from the warmed forward run:

```text
instructions:      -10.63%
cycles:            -7.08%
task clock:        -7.38%
branches:          -10.16%
branch misses:     -3.62%
cache references:  -3.12%
cache misses:      -3.52%
L1 loads:          -8.66%
L1 load misses:    -5.02%
dTLB loads:        -3.60%
dTLB load misses:  -4.08%
```

Allocation profile:

```text
same profile/output/thread shape
```

Heaptrack/Massif over three pairs:

```text
allocation calls:       -3,492,919 (-46.72%)
temporary allocations:    -650,325 (-73.87%)
heaptrack runtime:          -2.49s (-36.23%)
heaptrack peak heap: unchanged at 2 GiB
Massif peak total:       +21.31 MB (+1.09%)
Massif stack peak:       +48 bytes (+0.05%)
leaked bytes: unchanged
```

Raw allocation counts:

| metric | upstream | performance stack | delta |
|---|---:|---:|---:|
| allocation calls | 7,475,754 | 3,982,835 | -3,492,919 |
| temporary allocations | 880,362.7 | 230,037.3 | -650,325.3 |

## Possible Regressions

The changes are intended to preserve generated tile behavior. They mostly alter
temporary storage ownership, reserve sizes, and fast-path checks.

Potential risks:

- The sorted-node cache keeps a few decoded chunks per worker instead of one,
  which may retain a small amount of extra per-thread cache memory.
- Some geometry buffers now reserve or reuse capacity more deliberately. This
  reduces allocation churn, but retained capacity may slightly change heap
  shape.
- The clipping fast paths depend on existing bounds checks. Incorrect bounds
  handling would be a correctness bug, so those changes are kept narrow and
  only skip work when the existing geometry is already inside the tested box or
  no points need clipping for a given edge.

## Testing

Code checks:

```sh
git diff --check
```

Build:

```sh
cmake --build build --parallel 8
```

CTest:

```sh
ctest --test-dir build --output-on-failure
```

Result:

```text
No tests were found
```

### DIFF
--- a/include/coordinates_geom.h
+++ b/include/coordinates_geom.h
@@ -23,7 +23,7 @@ public:
 	TileBbox(TileCoordinates i, uint z, bool h, bool e);
 
 	std::pair<int,int> scaleLatpLon(double latp, double lon) const;
-	std::vector<Point> scaleRing(Ring const &src) const;
+	Ring scaleRing(Ring const &src) const;
 	MultiPolygon scaleGeometry(MultiPolygon const &src) const;
 	std::pair<double, double> floorLatpLon(double latp, double lon) const;
 
--- a/include/coordinates_geom.h
+++ b/include/coordinates_geom.h
@@ -23,7 +23,9 @@ public:
 	TileBbox(TileCoordinates i, uint z, bool h, bool e);
 
 	std::pair<int,int> scaleLatpLon(double latp, double lon) const;
+	void scaleRing(Ring &dst, Ring const &src) const;
 	Ring scaleRing(Ring const &src) const;
+	void scaleGeometry(MultiPolygon &dst, MultiPolygon const &src) const;
 	MultiPolygon scaleGeometry(MultiPolygon const &src) const;
 	std::pair<double, double> floorLatpLon(double latp, double lon) const;
 
--- a/include/geometry/correct.hpp
+++ b/include/geometry/correct.hpp
@@ -10,6 +10,7 @@
  * ----------------------------------------------------------------------------
  */
 
+#include <utility>
 #include <vector>
 #include <boost/geometry.hpp>
 #include <boost/geometry/geometries/point_xy.hpp>
@@ -25,7 +26,7 @@ namespace impl {
 template<typename C, typename T>
 static inline void result_combine(C &result, T &&new_element)
 {
-    result.push_back(new_element);
+    result.push_back(std::forward<T>(new_element));
 
    	for(std::size_t i = 0; i < result.size() - 1; ) {
         if(!boost::geometry::intersects(result[i], result.back())) {
--- a/include/geometry/correct.hpp
+++ b/include/geometry/correct.hpp
@@ -281,9 +281,11 @@ static inline std::vector<ring_t> correct(ring_t const &ring, boost::geometry::o
 	dissolve_find_intersections(new_ring, pseudo_vertices, start_keys);
 
 	if(start_keys.empty()) {
-		if(std::abs(boost::geometry::area(new_ring)) > remove_spike_min_area) 
-			return { new_ring };
-		else
+		if(std::abs(boost::geometry::area(new_ring)) > remove_spike_min_area) {
+			std::vector<ring_t> result;
+			result.push_back(std::move(new_ring));
+			return result;
+		} else
 			return { };
 	}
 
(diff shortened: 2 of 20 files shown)


Symmetricity:
I don't have a beefy machine to run this on to confirm the improvements spread to world generation - can someone test this to confirm the gains?

systemed:
This looks great - thank you! I'm away for a few days, but when I'm back I can run this over the planet and do before/after timings.

systemed:
Running the whole planet on my usual test machine with `--store` on SSD, current master:

With this PR:

	User time (seconds): 212929.53
	System time (seconds): 4173.74
	Percent of CPU this job got: 1059%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 5:41:37
	Maximum resident set size (kbytes): 131045104
	Major (requiring I/O) page faults: 65182734
	Minor (reclaiming a frame) page faults: 230944794
	Voluntary context switches: 107876668
	Involuntary context switches: 9592595
	File system inputs: 1343526800

With this PR:

	User time (seconds): 195730.78
	System time (seconds): 4369.10
	Percent of CPU this job got: 1031%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 5:23:13
	Maximum resident set size (kbytes): 126011668
	Major (requiring I/O) page faults: 64226308
	Minor (reclaiming a frame) page faults: 212481966
	Voluntary context switches: 140178142
	Involuntary context switches: 8720223
	File system inputs: 1234899512

So a very handy overall saving of 5.4%. I haven't yet sense-checked the output but will do that later.

systemed:
All seems to be good - duly merged. Thank you again.

PR #903, opened 2026-05-28
Fix UsedWays bounds check

Symmetricity:
This PR is AI generated.

## Summary

- fixes `UsedWays` bounds checks when the requested way id is exactly the
  current `usedList.size()`
- adds a small regression test for the first insert and exact-size growth cases

## Background

`UsedWays` stores relation-used way ids in a `std::vector<bool>`. Its existing
checks used `wayid > usedList.size()`, then indexed `usedList[wayid]`. When
`wayid == usedList.size()`, that is one past the end of the vector.

That can produce undefined behavior in both directions:

- `insert()` can write past the end instead of growing the vector first
- `at()` can read past the end instead of returning `false`

I checked the history around this code. The surrounding `UsedWays` storage was
introduced for memory savings, and later retained during relation-processing
changes, but I did not find a rationale for treating the exact-size boundary as
valid.

## Change

Use `>=` for the resize/read boundary so the vector grows before an exact-end
write, and exact-end reads are reported as absent.

This keeps the existing growth strategy and data structure unchanged.

## Expected Behavior

There should be no intended output change except avoiding undefined behavior at
the vector boundary. The exact-boundary case now grows by the same `+256`
increment that nearby out-of-range ids already used.

## Testing

- `git diff --check`
- `make test_osm_store`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,6 +149,7 @@ test: \
 	test_significant_tags \
 	test_sorted_node_store \
 	test_sorted_way_store \
+	test_osm_store \
 	test_tile_coordinates_set
 
 test_append_vector: \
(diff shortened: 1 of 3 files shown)


systemed:
Thank you!

PR #915, opened 2026-06-05
Read release version from VERSION

Symmetricity:
AI-generated PR.

## Summary

Use a checked-in `VERSION` file as tilemaker's release version source.

This updates CMake, Makefile builds, and Docker builds to read the same version
value instead of relying on `git describe` at build time. CI now treats
`VERSION` changes as runtime-impacting and validates that the file uses `X.Y.Z`
format.

Docker CI builds pass the same `VERSION` value into the Dockerfile so the final
image label is no longer stale or left as an unknown placeholder. They also
pass the Git-aware build version into CMake so the executable inside CI-built
Docker images remains traceable. Docker builds that do not pass that
build-version argument still compile with CMake's `v3.1.0+nogit` fallback.

For builds from a Git checkout, non-release-tag builds append commit metadata to
the displayed tilemaker version, for example `v3.1.0+g<commit>` or
`v3.1.0+g<commit>.dirty`. Clean exact builds from the matching release tag keep
the `v3.1.0` form. Builds where Git metadata is unavailable use
`v3.1.0+nogit`.

## Rationale

Release and package builds should not depend on tag lookup being available in a
particular checkout or build context. A single version file gives CMake,
Makefile, Docker, and release automation the same source of truth and makes
version bumps easier to review.

The optional Git suffix keeps development, CI, and custom-built binaries
traceable without requiring Git at runtime or changing release artifact version
strings.

## Testing

- `actionlint .github/workflows/ci.yml`
- CMake configure and build of `tilemaker`
- `tilemaker --help` reports a Git commit suffix for a non-tag Git checkout
- Building from a source tree without `.git` reports `v3.1.0+nogit`


Symmetricity:
This PR is part 1 of a series of 4 PRs focused on CI and release workflows. Each PR stacks on top of the previous, so I'll submit them one at a time.

The end state is a repeatable 'release' workflow that builds and publishes compiled release zips/deb files. Can be triggered manually by the maintainer or will automatically trigger when git tag with vX.Y.Z is added.

Symmetricity:
You can see an example of the end state here: https://github.com/Symmetricity/tilemaker/releases/tag/v3.1.0

systemed:
Looks good - thank you!

PR #918, opened 2026-06-17
Simplify buildings algorithm

systemed:
This adds a specialised simplification algorithm for buildings, which aims to preserve their rectilinear nature while reducing vertices. You can try it out using a layer config like this:

    "building": { "minzoom": 13, "maxzoom": 14, "combine_polygons_below": 15,
                  "simplify_algorithm": "buildings", "simplify_below": 15, "simplify_level": 0.0001 }

It works best with `combine_polygons_below`. Typical output is like this:

image

from this source data:

image

The algorithm was developed in Ruby, and ported to C++ with the help of Claude.

Performance is moderate - there are some opportunities for optimisation (such as replacing atan2 with a faster implementation), but given that the main impact is from `combine_polygons_below` anyway, this isn't a massive deal.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,6 +132,7 @@ file(GLOB tilemaker_src_files
 	src/shp_mem_tiles.cpp
 	src/shp_processor.cpp
 	src/significant_tags.cpp
+	src/simplify_buildings.cpp
 	src/sorted_node_store.cpp
 	src/sorted_way_store.cpp
 	src/tag_map.cpp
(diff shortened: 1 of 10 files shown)


systemed:
Additional commit now optimises union_many by using an rtree to only join adjoining polygons.

systemed:
Some very rough timings based on running Oxfordshire on an old Mac mini:

````
no combine, no simplify:  15.19 real   36.76 user   1.89 sys   22085632
no combine, simplify:     22.28 real   38.77 user   3.13 sys   22216704
old combine, no simplify: 65.96 real  129.22 user   3.95 sys   21233664
old combine, simplify:    67.17 real  127.34 user   4.15 sys   20840448
new combine, no simplify: 18.38 real   44.76 user   1.67 sys   21233664
new combine, simplify:    19.97 real   43.63 user   2.35 sys   20709376
````

Issue #867, opened 2025-12-19
Better error output


der-stefan:
When working on the JSON file, you sometimes make small mistakes like forgetting a comma, trying out comments (which don't work). All that tilemaker says is 
`Invalid JSON file.`

It would be helpful if there is a better error message that actually helps in debugging the style file.

etienneJr:
> trying out comments (which don't work)

tip: this is how I add comments in the json file
https://codeberg.org/cartes/serveur/src/commit/fc3bbff08bd2dbb6f886020984ca553f2679ae32/tilemaker/resources/config-openmaptiles.json#L16

<img width="403" height="124" alt="Image" src="https://github.com/user-attachments/assets/82cdf72d-0ae6-4ed3-8afc-343718877aa9" />

PR #919, opened 2026-06-20
Validate config JSON against schema

Symmetricity:
This PR was prepared with AI assistance.

Tilemaker now validates config JSON against a schema before reading required
settings from it.

The schema lives in `resources/config-schema.json` so it can be read, reviewed,
and reused by editors or external validation tools. The build embeds that schema
into `tilemaker`:

- CMake reads `resources/config-schema.json` and generates `config_schema.h` in
  the build directory.
- The Makefile generates `src/config_schema.h` from the same schema file.
- The generated Makefile header is ignored and removed by `make clean`.

This avoids a runtime dependency on finding a schema file while still keeping
the schema separate from the C++ code.

Why
---

Invalid config files could previously fail in two unhelpful ways:

- Syntax errors only reported `Invalid JSON file.`
- Structurally invalid but syntactically valid configs could reach
  `Config::readConfig`, where unchecked RapidJSON access could assert.

This specifically improves the cases reported in #867 and #875. For example,
the config attached to #875 is now rejected before `readConfig` with:

```
Invalid JSON file: missing required fields "compress", "name", "version", "description" at #/settings.
```

Other examples now include:

```
Invalid JSON file: Missing a comma or '}' after an object member. at offset 41.
```

```
Invalid JSON file: invalid type at #/layers/water/minzoom: expected "integer", got string.
```

Testing
---

- `make test_config_validator`
- `cmake -S . -B ... -DCMAKE_BUILD_TYPE=RelWithDebInfo`
- `cmake --build ... --parallel 8`
- smoke-tested malformed JSON, schema-invalid JSON, and the config from #875
- `make test`
- `git diff --check`


systemed:
Really good - thank you!

